PR #139, opened 2022-06-01
fix connect-web binary header test

cyinma:
@timostamm clarified on the `Uint8Array` [here](https://github.com/bufbuild/connect-crosstest/pull/138#pullrequestreview-992631411), this should fixed it


timostamm (approved):
Thank you!

PR #144, opened 2022-06-02
cleanup

cyinma:
fixes TCN-114


cmd/client/main.go
```diff
@@ -55,21 +64,28 @@ type flags struct {
...
+func bind(cmd *cobra.Command, flags *flags) error {
+	cmd.Flags().StringVar(&flags.host, hostFlagName, "127.0.0.1", "the host name of the test server")
+	cmd.Flags().StringVar(&flags.port, portFlagName, "", "the port of the test server")
+	cmd.Flags().StringVarP(
 		&flags.implementation,
-		"implementation",
+		implementationFlagName,
 		"i",
-		"connect",
+		"",
```
cyinma, line 88: remove default value as it is outdated and I cannot think of a good default, marked it as required instead

internal/crosstesting/crosstesting.go
```diff
@@ -16,9 +16,8 @@ package crosstesting
 
 import "testing"
 
-// TB is a trimmed down version of the standard library testing.TB interface.
-// connect-crosstest depends on this interface. All standard library *testing.T,
-// B, and F types implement this and can be used by connect-crosstest.
+// TB is a testing interface that connect-crosstest depends on. It is trimmed down
+// from the standard library testing.TB interface and added a successf method.
```
cyinma, line 20: not sure if this is a suitable comment

internal/interopgrpc/test_server.go
```diff
@@ -29,13 +29,15 @@ import (
...
 	testpb "github.com/bufbuild/connect-crosstest/internal/gen/proto/go/grpc/testing"
-	"github.com/bufbuild/connect-crosstest/internal/interopconnect"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
 
+// nonASCIIErrMsg is a non-ASCII error message.
+const nonASCIIErrMsg = "soirée 🎉" // readable non-ASCII
```
cyinma, line 39: cannot find of a good place to put this and share by both interopgrpc and interopconnect, considering that we wont actually have a lot of them, I think it might be better to just duplicate them

internal/crosstesting/crosstesting.go
doriable: The comment looks generally pretty good.

Nit: `%s/and added a successf/and adds a Successf`

doriable (approved):

PR #156, opened 2022-06-09
Remove gRPC and gRPC-web test cases against h3 server until trailer support is available

doriable:
Since we are adding strict reinforcement of `grpc-status` headers,
we cannot run gRPC and gRPC client tests against the h3 server
until trailer support is available.

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -91,7 +91,7 @@ func bind(cmd *cobra.Command, flags *flags) error {
 		"i",
 		"",
 		fmt.Sprintf(
-			"the client implementation tested, accepted values are %q, %q, %q, %q, %q, %q, %q, or %q",
+			"the client implementation tested, accepted values are %q, %q, %q, %q, %q, %q, %q, %q, or %q",
 			connectH1,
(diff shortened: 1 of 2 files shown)


cmd/client/main.go
```diff
@@ -212,15 +208,12 @@ func run(flags *flags) {
 		)
 	// For tests that depend on trailers, we only run them for HTTP2, since the HTTP3 client
 	// does not yet have trailers support https://github.com/lucas-clemente/quic-go/issues/2266
-	case connectH3, connectGRPCH3, connectGRPCWebH3:
+	// connectGRPCH3 and connectGRPCWebH3 have both been disabled since we are now strictly
+	// requiring `grpc-status` headers to be set on response, which requires trailer support.
+	// Once trailer support is available, they will be renabled.
```
timostamm, line 213: Do we need to disable gRPC-web? The protocol doesn't use HTTP trailers. It does have a concept of trailers, but they are encoded in the response body. 
As far as I see, the lack of trailer support in the H3 implementation is only an issue for gRPC.

PR #163, opened 2022-06-13
add server startup messages

cyinma:
fixes TCN-159

this added a message when server startup. instead of breaking the json stdout, I added a message field in the metadata. @timostamm can you confirm this works for you?

an example output:

```
❯ go run cmd/servergrpc/main.go --port 8123 --cert cert/localhost.crt --key cert/localhost.key
{"host":"localhost", "protocols":[{"protocol":"PROTOCOL_GRPC", "httpVersions":[{"major":2}], "port":"8123"}], "message":"server is listening on the above protocols."}
```


timostamm (commented):
As far as I see, this would change the following output:

```
CROSSTEST_VERSION=75774b9dabba88b201be341dd65dee928a252d49 docker-compose up
[+] Running 3/0
 ⠿ Network connect-web-main_default             Created                                                                                    0.0s
 ⠿ Container connect-web-main-server-grpc-1     Created                                                                                    0.0s
 ⠿ Container connect-web-main-server-connect-1  Created                                                                                    0.0s
Attaching to connect-web-main-server-connect-1, connect-web-main-server-grpc-1
connect-web-main-server-connect-1  | {"host":"localhost","protocols":[{"protocol":"PROTOCOL_GRPC_WEB","httpVersions":[{"major":1,"minor":1}],"port":"8080"},{"protocol":"PROTOCOL_GRPC_WEB","httpVersions":[{"major":2}],"port":"8081"},{"protocol":"PROTOCOL_GRPC","httpVersions":[{"major":2}],"port":"8081"}]}
connect-web-main-server-grpc-1     | {"host":"localhost", "protocols":[{"protocol":"PROTOCOL_GRPC", "httpVersions":[{"major":2}], "port":"8083"}]}
```

To contain `server is listening on the above protocols` twice, right? I'm not familiar with docker-compose. Would the message only be printed to standard out when _all_ servers are ready?

doriable:
I believe we want to print "server is listening to the above protocols" in a new line, not within the structure of the single print, right (basically output twice)?

timostamm:
For each server is not going to work :grimacing: We look for a message that indicates we can launch the test suite for _all_ servers.

cyinma:
> For each server is not going to work 😬 We look for a message that indicates we can launch the test suite for _all_ servers.

discussed with @doriable offline, and we think that in this case this will be something that we need to work on in connect-web docker-compose instead, therefore closing this

PR #169, opened 2022-06-15
update protobuf-es and connect-web

cyinma:


timostamm:
Jasmine continues to run specs even if an expectation failed. It's likely that the root cause is further up.

Did you [see](https://github.com/bufbuild/connect-web/pull/120):

> Most importantly, createConnectTransport() now creates a transport that uses the Connect protocol instead of gRPC-web. To use gRPC-web, switch to createGrpcWebTransport().

This might be the culprit.

cyinma:
> Jasmine continues to run specs even if an expectation failed. It's likely that the root cause is further up.
> 
> Did you [see](https://github.com/bufbuild/connect-web/pull/120):
> 
> > Most importantly, createConnectTransport() now creates a transport that uses the Connect protocol instead of gRPC-web. To use gRPC-web, switch to createGrpcWebTransport().
> 
> This might be the culprit.

yeh I just switched to it, and some of the errors disappeared, the remaining errors seems related to error details

cyinma:
> > Jasmine continues to run specs even if an expectation failed. It's likely that the root cause is further up.
> > Did you [see](https://github.com/bufbuild/connect-web/pull/120):
> > > Most importantly, createConnectTransport() now creates a transport that uses the Connect protocol instead of gRPC-web. To use gRPC-web, switch to createGrpcWebTransport().
> > 
> > 
> > This might be the culprit.
> 
> yeh I just switched to it, and some of the errors disappeared, the remaining errors seems related to error details

so seems like it need to be changed to use `rawDetails` for error details, I assume it is an intended changes? @timostamm

web/spec/connect-web.callback.spec.ts
```diff
@@ -330,10 +332,10 @@ describe("connect_web_callback_client", function () {
     });
     client.failUnaryCall({}, (err: ConnectError | undefined) => {
       expect(err).toBeInstanceOf(ConnectError);
-      expect(err?.code).toEqual(StatusCode.ResourceExhausted);
+      expect(err?.code).toEqual(Code.ResourceExhausted);
       expect(err?.rawMessage).toEqual("soirée 🎉");
       expect(err?.details.length).toEqual(1);
-      expect(err?.details[0].equals(Any.pack(expectedErrorDetail))).toBeTrue();
+      expect(err?.details[0].equals(expectedErrorDetail)).toBeTrue();
```
timostamm, line 338: I think this would be slightly more robust if we called the equals method on the known good value:

```suggestion
      expect(expectedErrorDetail.equals(err?.details[0])).toBeTrue();
```

timostamm (approved):

PR #238, opened 2022-08-09
test servers health check on spin up and envoy depends on test servers

cyinma:
this minimized the chance of web client test getting flaky no healthy upstream error, the issue is that docker-compose `depends_on` config doesn't guarantee the server is ready for request. It wasn't an issue when the tests of go and web are running together as the go tests were run first, so by the time the web tests start all server should already be ready for request. Attempted multiple way (e.g check connection before starting the test, setup retry policy in envoy) and this seems to be the most reasonable way to do this for now.

update:
turns out there is also an issue within envoy, where the envoy didn't depends on the connect and grpc test server, so the dns resolution for the server that not spin up yet will fail, and when we spin up the test server and start the test, the dns resolution might not be refreshed yet as it has a 5000ms refresh rate. updated envoy to depends on both servers, and the test clients only depends on envoy

### DIFF
--- a/Dockerfile.crosstest
+++ b/Dockerfile.crosstest
@@ -19,10 +19,14 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 RUN --mount=type=cache,target=/root/.cache/go-build \
     CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
     go build -o /go/bin/servergrpc ./cmd/servergrpc
+RUN --mount=type=cache,target=/root/.cache/go-build \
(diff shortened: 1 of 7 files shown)


pkwarren:
One other option is to use the long format `depends_on` with `service_healthy` (and define a healthcheck for each service that other services depend on): https://github.com/compose-spec/compose-spec/blob/master/spec.md#long-syntax-1

cyinma:
> One other option is to use the long format `depends_on` with `service_healthy` (and define a healthcheck for each service that other services depend on): https://github.com/compose-spec/compose-spec/blob/master/spec.md#long-syntax-1

updated, this looks much better, thanks for the suggestion!

doriable (approved):

pkwarren (approved):

PR #262, opened 2022-08-29
update dependencies

cyinma:


stefanvanburen (approved):
should we also bump the go version in the Dockerfile (for https://github.com/bufbuild/connect-crosstest/pull/232)?

cyinma:
> should we also bump the go version in the Dockerfile (for #232)?

not sure the originally intention of skipping that update, will sync with @doriable later, but that should be in a separated PR anyway

PR #330, opened 2022-11-30
upgrade dependencies and remove tag of alpine docker image to use latest

cyinma:


akshayjshah (approved):

Dockerfile.crosstest
```diff
@@ -20,7 +20,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
     go build -o /go/bin/servergrpc ./cmd/servergrpc
 
-FROM --platform=${TARGETPLATFORM} alpine:3.16.3
```
akshayjshah, line 23: If you want, I think we're okay to just make this `alpine` and use latest.

PR #332, opened 2022-11-30
add connect grpc-web client to envoy server tests

cyinma:
while upgrading the dependencies, I noticed that we don't have tests for connect-go's client with grpc-web protocol to connect and grpc server via envoy 😐


doriable (approved):
🙏🏼

PR #386, opened 2023-02-15
Add test case fail_server_streaming_after_response

timostamm:
gRPC-web usually appends `grpc-status` and related fields to the response body. But if a response does not contain any messages, the server may set the fields as response headers instead. This is called a trailers-only response.

Currently, `fail_server_streaming` is running clients against the trailers-only response, but not against fields appended to the response body. This adds a new test for this case:

> Client calls `FailStreamingOutputCall`, and asks for four response messages. The server
responds with the response messages, then an error with status `RESOURCE_EXHAUSTED`
and a non-ASCII message with error details. The client verifies that four response messages
and the error status with code, message, and details was received.

Instead of adding a separate RPC, we update the existing implementation of `rpc FailStreamingOutputCall` to make use of an [existing request message field](https://github.com/bufbuild/connect-crosstest/blob/ea9e239e67e3f95e08c6fd41750448569b484d05/internal/proto/grpc/testing/messages.proto#L171):

```proto
  // Configuration for each expected response message.
  repeated ResponseParameters response_parameters = 2;
```

If the client asks for response messages via this field, we can be certain to receive the status fields appended to the body. If the client leaves the field unset, the server may reply with trailers-only.

Note that this only updates web clients, not the connect-go and grpc-go clients.


timostamm:
[We could make use of ](https://github.com/bufbuild/connect-crosstest/pull/386/commits/98d51776a0ba1b40beadaa2f83c707aa44b4cf81) updates the Go clients to retain behaviour with the updated RPC, but I'd still recommend to follow-up with an implementation of `fail_server_streaming_after_response`.

akshayjshah:
Hey @timostamm and @joshcarp (possibly @smaye81 too?) - can we get this polished up and merged in?

timostamm:
This passes `make all dockercomposetest` after merging in main. Pinging @cyinma and @doriable to make sure I'm pressing the right buttons.

cyinma:
> This passes `make all dockercomposetest` after merging in main. Pinging @cyinma and @doriable to make sure I'm pressing the right buttons.

yes thats correct, I will defer the actual review to @joshcarp and @smaye81

PR #387, opened 2023-02-16
Add go tests for fail_server_streaming_after_response

joshcarp:


internal/interop/interopconnect/test_cases.go
```diff
@@ -657,6 +657,39 @@ func DoFailServerStreamingWithNonASCIIError(t crosstesting.TB, client connectpb.
...
+			Size: int32(s),
+		}
+	}
+	req := &testpb.StreamingOutputCallRequest{
+		ResponseType:       testpb.PayloadType_COMPRESSABLE,
+		ResponseParameters: respParam,
+	}
+	stream, err := client.FailStreamingOutputCall(context.Background(), connect.NewRequest(req))
+	require.NoError(t, err)
+	for i := 0; i < 4; i++ {
```
timostamm, line 673: What do you think about doing:

```suggestion
	for i := 0; i < len(respSizes); i++ {
```

? We are asking the server for that many responses, so it seems fair to use exactly the same source of truth to assert the actual number of responses we receive.

internal/interop/interopgrpc/test_cases.go
```diff
@@ -550,6 +550,42 @@ func DoFailServerStreamingWithNonASCIIError(t crosstesting.TB, client testpb.Tes
...
+			Size: int32(s),
+		}
+	}
+	req := &testpb.StreamingOutputCallRequest{
+		ResponseType:       testpb.PayloadType_COMPRESSABLE,
+		ResponseParameters: respParam,
+	}
+	stream, err := client.FailStreamingOutputCall(context.Background(), req, args...)
+	require.NoError(t, err)
+	for i := 0; i < 4; i++ {
```
timostamm, line 567: And here as well?

joshcarp:
```suggestion
	for i := 0; i < len(respSizes); i++ {
```

akshayjshah (approved):

PR #434, opened 2023-04-19
Add idempotency rpcs

buildbreaker:


timostamm:
Hey @buildbreaker, did you see the `rpc CacheableUnaryCall`? What do you think about setting `idempotency_level = NO_SIDE_EFFECTS` there, and adding a `rpc CacheableStreamingOutputCall`? My thinking is that it will be easier to cover all code paths when we keep the RPCs separate.

buildbreaker:
@timostamm That sounds like a better plan. I can make the changes here.

jchadwick-buf:
Superseded in #441.

PR #445, opened 2023-05-05
Add missing CORS exposed headers for GET

jchadwick-buf:
This is breaking my tests right now for `connect-es` cacheable unary support.


timostamm (approved):
I was thinking about the CORS setup in Envoy in https://github.com/bufbuild/connect-crosstest/pull/441, but totally missed this one 🙈

jchadwick-buf:
> I was thinking about the CORS setup in Envoy in #441, but totally missed this one 🙈

Wait a minute, that's a good point. I should add these to the headers exposed by Envoy as well, for good measure.

PR #469, opened 2023-06-14
Add additional check for Fetch API duplicate headers

smaye81:
This adds an additional check for any duplicate headers that may have been set by the Fetch API. The `Headers` object will concatenate duplicate header values into one comma-delimited string. As a result, this was causing the tests to fail for any client hitting a Node server using this `Headers` object.

### DIFF
--- a/internal/interop/interopconnect/test_cases.go
+++ b/internal/interop/interopconnect/test_cases.go
@@ -298,6 +298,9 @@ func validateMetadata(
 	expectedStringHeaders map[string][]string,
 	expectedBinaryHeaders map[string][][]byte,
 ) {
+	fmt.Println("Using a canonical header: ", trailer.Values("X-Grpc-Test-Echo-Trailing-Bin"))
+	fmt.Println("Using all lowercase: ", trailer.Values("x-grpc-test-echo-trailing-bin"))
+	fmt.Println("Using the map directly: ", trailer["x-grpc-test-echo-trailing-bin"])
 	for key, values := range expectedStringHeaders {
 		actualValues := header.Values(key)
 		// If the returned header values are not equal to what we expect, next check whether the header
@@ -318,11 +321,6 @@ func validateMetadata(
 		}
 	}
 	for key, values := range expectedBinaryHeaders {
-		fmt.Println("expected trailers key ", key)
-		fmt.Println("expected trailer values ", values)
-		fmt.Println("expected trailer values len", len(values))
-		fmt.Println("actual trailer values ", trailer.Values(key))
-		fmt.Println("actual trailer values len", len(trailer.Values(key)))
 		actualValues := trailer.Values(key)
 		// If the returned trailer values are not equal to what we expect, next check whether the trailer
 		// values were combined into one comma-delimited string. The Fetch API Headers object will return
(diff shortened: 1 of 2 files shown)


internal/interop/interopconnect/test_cases.go
```diff
@@ -374,7 +390,7 @@ func DoDuplicatedCustomMetadataUnary(t crosstesting.TB, client connectpb.TestSer
 		t,
 		client,
 		map[string][]string{
-			leadingMetadataKey: {leadingMetadataValue, leadingMetadataValue + ",more_stuff"},
+			leadingMetadataKey: {leadingMetadataValue, leadingMetadataValue + ";more_stuff"},
```
smaye81, line 393: Changed to a semicolon so that the tests didn't break, since it would split on this comma also. Note that in the wild, it is still possible to use commas in duplicate header values, but this is to make our testing lives easier.

internal/interop/interopconnect/test_cases.go
```diff
@@ -298,23 +299,38 @@ func validateMetadata(
...
 			assert.True(t, ok)
 		}
 	}
 	for key, values := range expectedBinaryHeaders {
-		assert.Equal(t, len(values), len(trailer.Values(key)))
+		actualValues := trailer.Values(key)
+		// If the returned trailer values are not equal to what we expect, next check whether the trailer
+		// values were combined into one comma-delimited string. The Fetch API Headers object will return
+		// duplicate headers this way.
+		if len(values) != len(actualValues) && len(actualValues) == 1 {
```
timostamm, line 325:
```suggestion
		// The server may have combined multiple lines for a field to a single line, see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3
```

internal/interop/interopconnect/test_cases.go
```diff
@@ -298,23 +299,38 @@ func validateMetadata(
...
 ) {
 	for key, values := range expectedStringHeaders {
-		assert.Equal(t, len(values), len(header.Values(key)))
+		actualValues := header.Values(key)
+		// If the returned header values are not equal to what we expect, next check whether the header
+		// values were combined into one comma-delimited string. The Fetch API Headers object will return
+		// duplicate headers this way.
+		if len(values) != len(actualValues) && len(actualValues) == 1 {
+			actualValues = strings.Split(actualValues[0], ", ")
+		}
```
timostamm, line 308: We're not covering all cases here. All following forms are semantically equivalent:
```
Foo: a
Foo: b
Foo: c
Foo: d
```

```
Foo: a, b, c, d
```

```
Foo: a
Foo: b, c, d
```

We need a comment that this is not a complete semantic comparison.

smaye81: Would that third case ever occur though? I guess if a non-Fetch API server decided to concatenate values on their own and add them as a duplicate?

I added a comment too.  Lmk what you think.

timostamm: I don't know if that will ever occur. It would be best if validateMetadata() did a semantical comparison, so that this uncertainty is removed. Doing so will end up as a significant piece of code that needs test coverage on its own - so the status quo with a comment is acceptable in my book.

timostamm (approved):

PR #471, opened 2023-06-19
Add ability to start insecure servers

smaye81:
This adds the ability to start an insecure local Connect server. By providing the new `insecure` flag, it will start a Connect server which serves cleartext over the specified HTTP/1.1 and HTTP/2 ports. Note that HTTP/3 ignores this flag.

This is useful for developing new clients and servers for new language implementations locally. It allows you to easily spin both up without the ceremony of creating certificates. Some sample commands:

```bash
# Start a local connect server serving cleartext on 3000 and 3001
go run cmd/serverconnect/main.go --h1port 3000 --h2port 3001 --insecure --h3port 3002

# Start a local connect server serving TLS on all ports
go run -v cmd/serverconnect/main.go --h1port 3000 --h2port 3001 --h3port 3002 --cert "cert/server-connect.crt" --key "cert/server-connect.key"
```

Note: This does not affect the gRPC server. That can be a follow-up PR if desired.

### DIFF
--- a/cmd/serverconnect/main.go
+++ b/cmd/serverconnect/main.go
@@ -246,33 +246,33 @@ func run(flags *flags) {
 	}
 }
 
-func newH1Server(f *flags, handler http.Handler) http.Server {
-	h1Server := http.Server{
-		Addr:    ":" + f.h1Port,
+func newH1Server(flags *flags, handler http.Handler) *http.Server {
+	h1Server := &http.Server{
+		Addr:    ":" + flags.h1Port,
 		Handler: handler,
 	}
-	if !f.insecure {
-		h1Server.TLSConfig = newTLSConfig(f.certFile, f.keyFile)
+	if !flags.insecure {
+		h1Server.TLSConfig = newTLSConfig(flags.certFile, flags.keyFile)
 	}
 	return h1Server
 }
 
-func newH2Server(f *flags, handler http.Handler) http.Server {
-	h2Server := http.Server{
-		Addr:    ":" + f.h2Port,
+func newH2Server(flags *flags, handler http.Handler) *http.Server {
+	h2Server := &http.Server{
+		Addr:    ":" + flags.h2Port,
 		Handler: handler,
 	}
-	if !f.insecure {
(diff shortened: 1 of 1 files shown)


cmd/serverconnect/main.go
```diff
@@ -228,6 +246,36 @@ func run(flags *flags) {
...
+		Handler: handler,
+	}
+	if !f.insecure {
+		h1Server.TLSConfig = newTLSConfig(f.certFile, f.keyFile)
+	}
+	return h1Server
+}
+
+func newH2Server(f *flags, handler http.Handler) http.Server {
+	h2Server := http.Server{
```
smaye81, line 261: Not sure if this should be `http2.Server`. The previous code uses `http`, so I kept it the same.

cmd/serverconnect/main.go
```diff
@@ -73,7 +88,8 @@ func bind(cmd *cobra.Command, flagset *flags) error {
 	cmd.Flags().StringVar(&flagset.h3Port, h3PortFlagName, "", "port for HTTP/3 traffic")
 	cmd.Flags().StringVar(&flagset.certFile, certFlagName, "", "path to the TLS cert file")
 	cmd.Flags().StringVar(&flagset.keyFile, keyFlagName, "", "path to the TLS key file")
-	for _, requiredFlag := range []string{h1PortFlagName, h2PortFlagName, certFlagName, keyFlagName} {
+	cmd.Flags().BoolVar(&flagset.insecure, insecureFlagName, false, "whether to serve cleartext or TLS. Note that the HTTP/3 port will always serve TLS. ")
```
timostamm, line 91: Would it be more clear with

```suggestion
	cmd.Flags().BoolVar(&flagset.insecure, insecureFlagName, false, "whether to serve cleartext or TLS. HTTP/3 requires TLS.")
```

?

cmd/serverconnect/main.go
smaye81: Updated to use `h2c.NewHandler`. I tested this locally with an insecure client using `connect-h2` and `insecure` as the implementation and it worked. Also tested `connect-h1` and `insecure`, which also worked.

timostamm (approved):

PR #477, opened 2023-06-20
Add connect-node-fastify server

smaye81:
This adds a new Connect server using `connect-fastify` and `connect-node`. It wires it into the docker-compose setup and allows for serving HTTP/1.1 and HTTP/2 traffic using cleartext and TLS. Fastify does not offer HTTP/3 support as of yet, so that path is not available. Note: this does not yet use TLS, but can be added as a follow-up PR.

In addition, this also refactors a few things such as:
* adds TypeScript support to the web portion
* renames images, Dockerfiles, etc. to match the language so that adding new languages can follow one pattern. For example, `client-connect-to-server-connect-go-h2` becomes `client-connect-go-to-server-connect-go-h2` and uses `Dockerfile.crosstestgo` instead of `Dockerfile.crosstest`.
* updates to connect-go v1.9.0 for https://github.com/bufbuild/connect-go/pull/528 which fixes an issue with parsing response headers from a connect-node server.
* updates to connect-es v0.11.0 for https://github.com/bufbuild/connect-es/pull/695, which fixes an issue with parsing response bodies for non-JSON content types.


internal/interop/interopconnect/test_cases.go
```diff
@@ -501,7 +501,7 @@ func customMetadataServerStreamingTest(
 		require.NoError(t, stream.Err())
 	}
 	assert.NoError(t, stream.Close())
-	validateMetadata(t, stream.ResponseHeader(), stream.ResponseTrailer(), customMetadataString, customMetadataBinary)
+	// validateMetadata(t, stream.ResponseHeader(), stream.ResponseTrailer(), customMetadataString, customMetadataBinary)
```
smaye81, line 504: Temporary until this is fixed in connect-go (see https://github.com/bufbuild/connect-go/issues/524). Won't land this PR until this can be uncommented.

docker-compose.yaml
```diff
@@ -31,124 +42,158 @@ services:
...
+        TEST_CONNECT_GO_BRANCH: "${TEST_CONNECT_GO_BRANCH:-}"
+    command: >
+        /bin/sh -c "
+        echo Waiting for Fastify to start...;
+        while ! nc -z server-connect-node-fastify 8084;
+        do
+          sleep 1;
+        done;
+        /usr/local/bin/client --host='server-connect-node-fastify' --port='8084' --implementation='connect-h1' --insecure
+        "
```
smaye81, line 86: Super hacky, but the tests seem to fail unless we add some sort of wait to the fastify container. The `depends-on` attribute waits until the container is up, but it seems it takes a bit longer for Fastify to bind to the port. Without something like this, we get a `Connection refused` on the container. Open to suggestions for how to fix this better.

timostamm: What about [healthcheck](https://docs.docker.com/compose/compose-file/05-services/#healthcheck)?

web/server/fastify/server.ts
```diff
@@ -0,0 +1,98 @@
...
+const HOST = "0.0.0.0";
+
+const CORS_OPTIONS = {
+  // Reflects the request origin. This should only be used for development.
+  // Production should explicitly specify an origin
+  origin: true,
+  methods: [...connectCors.allowedMethods],
+  allowedHeaders: [...connectCors.allowedHeaders],
+  exposedHeaders: [...connectCors.exposedHeaders],
+};
```
timostamm, line 38: You have to include the application-specific headers here. See https://github.com/bufbuild/connect-es/blob/b6187e9b2007bf178c0a6dea79cab3f30ff09c1a/packages/connect-node-test/src/helpers/testserver.ts#L177 for reference.

Makefile
```diff
@@ -81,30 +81,32 @@ checkgenerate:
...
-	docker-compose run client-connect-grpc-web-to-envoy-server-connect-h1
-	docker-compose run client-connect-grpc-web-to-envoy-server-grpc-h1
-	docker-compose run client-connect-grpc-to-server-grpc
-	docker-compose run client-grpc-to-server-connect
-	docker-compose run client-grpc-to-server-grpc
+	docker-compose run client-connect-go-to-server-connect-go-h1
+	docker-compose run client-connect-go-to-server-connect-go-h2
+	docker-compose run client-connect-go-to-server-connect-go-h3
+	docker-compose run client-connect-go-to-server-connect-node-fastify-h1
+	docker-compose run client-connect-go-to-server-connect-node-fastify-h2
```
timostamm, line 88: We're running connect-go with the Connect protocol against the connect-es fastify server over HTTP/1.1 and HTTP/2, both cleartext. Nice!

Can we also run:
- connect-go with the gRPC protocol over H1
- connect-go with the gRPC protocol over H2
- connect-go with the gRPC-web protocol over H1
- connect-go with the gRPC-web protocol over H2
- grpc-go (always the gRPC protocol, always H2)

This will give us good coverage over most code paths in the connect-es server.

Makefile
```diff
@@ -81,30 +81,32 @@ checkgenerate:
...
 .PHONY: dockercomposetestweb
 dockercomposetestweb: dockercomposeclean
-	docker-compose run client-web-connect-web-to-server-connect-h1
-	docker-compose run client-web-connect-grpc-web-to-server-connect-h1
-	docker-compose run client-web-connect-grpc-web-to-envoy-server-connect
-	docker-compose run client-web-connect-grpc-web-to-envoy-server-grpc
-	docker-compose run client-web-grpc-web-to-server-connect-h1
-	docker-compose run client-web-grpc-web-to-envoy-server-connect
-	docker-compose run client-web-grpc-web-to-envoy-server-grpc
+	docker-compose run client-connect-web-to-server-connect-go-h1
```
timostamm, line 103: Can we run all browser clients against the connect-es server as well?

web/server/fastify/program.ts
```diff
@@ -0,0 +1,45 @@
...
+
+function validateNumber(value: string) {
+  const parsedValue = parseInt(value, 10);
+  if (Number.isNaN(value)) {
+    throw new InvalidArgumentError("option must be a number.");
+  }
+  return parsedValue;
+}
+
+program
```
timostamm, line 15: The go `serverconnect` and `servergrpc` commands print [server.v1.ServerMetadata](https://github.com/bufbuild/connect-crosstest/blob/0a1902575a342e9cbd62453bb7f832179cff5d11/proto/server/v1/server.proto#L19) at startup. 

Is the output parsed anywhere? :thinking:

docker-compose.yaml
```diff
@@ -188,7 +233,7 @@ services:
     entrypoint: npm run test -- --docker --host="envoy" --port="9092" --implementation="connect-grpc-web"
     depends_on:
       - envoy
-  client-web-grpc-web-to-server-connect-h1:
+  client-connect-web-grpc-web-to-server-connect-go-h1:
```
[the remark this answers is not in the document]
smaye81: OK, makes sense. I still think that having `web` right after `client-` is a bit redundant, at least in this case, but I'll leave it until we can organize things even further later.

Expanding on what you mentioned above and what I had in this PR, I think a good template is:

`client-{client implementation / protocol}-to-server-{server implementation}-{http version}`

where `client implementation / protocol` would be one of:
```
grpc-go (gRPC Go client)
grpc-web (gRPC-Web client)
connect-web-connect (Connect-Web client using Connect protocol)
connect-web-grpc-web (Connect-Web client using gRPC-Web protocol)
```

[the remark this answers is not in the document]
smaye81: Actually, the more I think about, I'd really like to change these now if possible. It makes things a lot easier to reason about now that we have multiple clients and servers. Would you be OK with making those changes as long as I fixed your issue with the misleading grpc-web names?

web/server/fastify/program.ts
smaye81: It doesn't appear to be used anywhere, but I use that same `ServerMetadata` object to print out the information in the server startup for Fastify if you like?

docker-compose.yaml
smaye81: I looked at that actually, but I need to test multiple ports (`8084` and `8085`) and there's no way to test multiple ports in docker-compose (or in the Dockerfile). I could create two containers, one for each port but that seemed like overkill.

timostamm: You don't need to listen for multiple ports if they are bound in sequence (they are). Here is an example: https://github.com/bufbuild/connect-crosstest/pull/479/files

The important part is that the health check is defined on the service we want to check, not on multiple clients. This is a definite improvement in my book.

I actually think it would be preferable to treat each run of the test suite as a discrete unit. That also means listening to only one port from a process. Out of scope for this PR IMO.

docker-compose.yaml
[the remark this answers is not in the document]
timostamm: The names are misleading, but there is also a duplicate service name. Apparently, docker does not care, but I'm sure one of the test suites isn't even running. We really have to be careful with the names. Yes, I'm OK with making the change, but please be careful, and coordinate changes with https://github.com/bufbuild/connect-crosstest/pull/478

web/server/fastify/program.ts
timostamm: Yes, let's stick with the pattern.

Makefile
smaye81: Added browser clients against connect-node. I didn't do the envoy path, though. Not sure if we should?

buf.work.yaml
```diff
@@ -1,3 +1,3 @@
 version: v1
 directories:
-  - internal/proto
+  - proto
```
timostamm, line 3: It's a reasonable change, but we have at least one command using the old path [here](https://github.com/bufbuild/connect-es/blob/4df67c9b8bda28f2bc52a7c6aa8d901b02607cdb/Makefile#L133). I'm not sure what Kotlin and Swift do.

web/dist/gen/proto/connect-web/grpc/testing/empty_pb.js
```diff
@@ -0,0 +1,44 @@
+var __defProp = Object.defineProperty;
```
timostamm, line 1: No files in dist should be committed.

web/server/build.js
```diff
@@ -0,0 +1,14 @@
...
+// because setting to module will cause grpc-web code to break.
+//
+// In addition, our esbuild is done via the JavaScript API so that we can use glob syntax
+(async () => {
+  let result = await esbuild.build({
+    entryPoints: await glob(["server/**/*.ts", "gen/**/*.ts"]),
+    outdir: "dist",
+    format: "cjs",
+  });
+})();
```
timostamm, line 14: This seems complicated. Can we not simply start the server with `npx tsx ...`?

web/server/fastify/server.ts
```diff
@@ -0,0 +1,106 @@
...
+  const h2Server = createH2Server(opts);
+  await h2Server.register(fastifyCors, interop.corsOptions);
+  await h2Server.register(fastifyConnectPlugin, { routes });
+  await h2Server.listen({ host: HOST, port: opts.h2port });
+
+  const serverData = getServerMetadata(opts);
+  console.log(serverData.toJsonString());
+  return new Promise<void>((resolve) => {
+    resolve();
+  });
```
timostamm, line 105:
```suggestion
```

Makefile
timostamm: Yes we should, but leave it up to you whether it's the right time.

docker-compose.yaml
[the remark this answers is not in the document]
smaye81: Hopefully I have them sorted out. I also left a comment on #478 to request following the same naming convention.

web/server/build.js
smaye81: I tried that at first and ran into the `incremental` breaking change with esbuild v0.17 and originally decided to punt on it until I could remove Karma. But, point taken. I pinned esbuild to v0.16.17 instead and removed all the complications around building.  Still want to remove Karma, but this is a much better interim solution.

buf.work.yaml
smaye81: I can fix that path once this lands. I checked both repos and they do not seem to rely on that path. Also followed up with Alan and Rebello to verify.

srikrsna-buf (approved):

PR #485, opened 2023-06-30
Fix push for Dockerfile.crosstestweb

smaye81:
This fixes the push step for the Dockerfile.crosstestweb image as that image does not require a multi-platform build.


cyinma (approved):
lgtm

PR #508, opened 2023-07-24
Add connect.conformance protos and c++/grpc code gen

Alfus:
Which are wire-compatible copies of the exiting protos, in connect-specific namespace/package.
I am including an empty gRPC c++ cross test client runner, to verify it compiles.
The c++ setup matches protovalidate-cc (which is modeled after envoy-proxy)
image


timostamm:
Is there a benefit to adding a copy? The messages might be wire-compatible, but the RPCs are not. If we want to change the namespaces, I think it would be more straight-forward in one swoop. There a several implementations using the proto files.

Alfus:
@timostamm This lets us migrate incrementally. My next PR adds support for both to the go connect server, unblocking c++.

Alfus:
I am going to split out the proto change.

akshayjshah (commented):
Something seems strange about the generated code added in this PR - can we keep _all_ generated Go code in `internal/gen` and all generated TS in `web/gen`? Ideally the generated cpp would also live in a single directory.

PR #510, opened 2023-07-25
Move test protos into `connectrpc.conformance` package

Alfus:
So they don't conflict with the grpc c++ library.


timostamm (commented):
> So they don't conflict with the grpc c++ library.

Ah. Are there global registry conflicts because grpc c++ compiles in grpc/testing/test.proto?

The downside of the name change is that our servers and clients are no longer compatible with [grpc interop tests](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md). They were never fully compatible in the first place because we don't support all ecosystem features and because we added a couple of RPCs. But the change means we fully deviate. 

IMO, adding grpc c++ gives us good bump in confidence, since our tests never ran against the full matrix of grpc implementations before. And the namespace change can be a clear cut to potentially evolve the test suite. So I'm personally on board, I just want to make sure this what we want.

connect-kotlin, connect-swift, and connect-es use the docker images published by this repository, so as a follow-up, the proto files that the repositories vendor need to be updated, and the client/server implementations will need some minor updates.

Alfus:
Yes, there is a global registry. I could try to use the versions shipped with grpc for c++

Alfus:
@timostamm It looks like these protos have already been changed in a number of ways from the original. See https://github.com/bufbuild/connect-crosstest/compare/alfus/grpcclient?expand=1 where I had to comment out a number of tests.

timostamm (approved):
👍 for the package name change and making a clear cut from the gRPC test protos.

Could you add tickets to update connect-kotlin, connect-swift, and connect-es?

akshayjshah:
LGTM to me too. Once we're _all_ the way through refactoring in here, we should definitely update the other implementations as Timo suggests.

@timostamm AFAIK the problem wasn't the global registry - it's that you can't compile C++ using the namespace "connect" on Macs b/c it conflicts with some other built-in identifier.

akshayjshah (approved):

PR #512, opened 2023-07-28
Add grpc c++ client conformance tests

Alfus:
The minimal footprint to get a grpc-based binary that can run the conformance tests. 
To test run:
 - run a test server, for example `go run cmd/serverconnect/main.go --insecure --h1port=8080 --h2port=8081`
 - run the conformance tests `PORT=8081 bazel run //cc:grpc_client_test`

Will require follow up PRs to fully integrate into crosstest docker images and CI.

Also note that the grpc client does not properly convert 404's into UNIMPLEMENTED, instead returning UNKNOWN.


cc/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//cc:__subpackages__"])
+
+cc_test(
+    name = "grpc_client_test",
+    srcs = ["grpc_client_test.cc"],
+    env = {"PORT": "8081"},
```
timostamm, line 20: Just FYI, this seems to override env variables.

timostamm (approved):
Mostly passes against the Node.js server too, nice.

Against that, duplicate metadata tests fail with:

```
Value of: actualValues
Expected: has 2 elements and there exists some permutation of elements such that:
 - element #0 is equal to "\n\v\n\v\n\v", and
 - element #1 is equal to "\n\v\n\v\n\v\n"
  Actual: { "" }, which has 1 element
```

Ran with node v18:

```bash
cd web
npm ci
npx tsx server/fastify/program.ts start --h1port=8086 --h2port=8081 --insecure
```


The Node.js server is the one that combines header lines, perhaps something is off in the comparison? We can sort it out when integrating the client...

cc/BUILD.bazel
Alfus: Last I checked bazel ignored these when you use 'bazel run' , and uses them when you run 'bazel test' (which in turn ignores env)

timostamm: Here's what I saw:

```bash
$ bazel --version
6.3.0
$ PORT=9999 bazel run //cc:grpc_client_test
...
14: failed to connect to all addresses; last error: UNAVAILABLE: ipv4:127.0.0.1:8081: Socket closed
...
```

Just pointing out for anyone who might want to run it.

Alfus: Thanks for letting me know! (maybe the fixed it)

PR #513, opened 2023-08-02
Add bazel targets for test protos

Alfus:
So they can be referenced by other repos that use bazel.


timostamm (approved):
Is it common to have the bazel file deep in the tree, instead of next to buf.yaml?

Sounds useful either way.

Alfus:
The BAZEL.build files are supposed to be as close to the source as possible, it's only when bootstrapping non-bazel repos, that the BAZEL.build files end up in the root.
This matches what we ended up doing for protovalidate: https://github.com/bufbuild/protovalidate/tree/main/proto/protovalidate

PR #535, opened 2023-08-24
Add c++ image with client tests and Docker Compose services

doriable:
This adds a new Docker image for the C++ tests, Docker Compose services that run the
tests cases, a make target for them, and an on-demand Github Action.
There are two cases added to Docker Compose, one testing h2 without tls and the other with h2 + TLS.
The Docker image with Bazel is not the most performant when run locally, so I would
recommend running the jobs mostly in CI.

### DIFF
--- a/.github/workflows/crosstest-cc.yaml
+++ b/.github/workflows/crosstest-cc.yaml
@@ -1,0 +1,23 @@
+name: crosstest-cc
+on:
+  workflow_dispatch
+permissions:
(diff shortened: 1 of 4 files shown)


Alfus:
grpc only supports http/2.0 with and without tls

doriable:
Updated test cases + PR description to test h2 with and without tls.

smaye81 (approved):

PR #538, opened 2023-08-28
Distinguish between HTTP trailers and application trailers

jhump:
For gRPC, there is no difference since application trailers are transmitted via HTTP trailers.

But for Connect and gRPC-Web, application trailers go in the body, not in HTTP trailers.

This verifies that the server protocol is putting trailing metadata where it's expected: in HTTP trailers for gRPC and elsewhere for Connect and gRPC-Web.


akshayjshah (approved):
LGTM - thanks for this, Josh!

PR #540, opened 2023-09-01
Run make to update files

smaye81:
This runs `make` on the repo to generate files and run the license header.

Note that this renames `Dockerfile.cc` to `Dockerfile.crosstestcc` to prevent `license-header` from adding a header to the Dockerfile.


smaye81:
@jhump @Alfus was doing some dependabot updates and noticed a few files being changed when I ran `make`. Putting these files in their own PR for your review.

jhump (approved):

PR #591, opened 2023-10-11
Refine server proto definition

smaye81:
This makes some additional refinements to the conformance server definition.

### DIFF
--- a/proto/connectrpc/conformance/v1alpha1/service.proto
+++ b/proto/connectrpc/conformance/v1alpha1/service.proto
@@ -73,25 +73,12 @@ service ConformanceService {
   // If the input stream is empty, the server should send a single response
   // message that includes no data and only the request properties (headers,
   // timeout).
-  //
-  // Note that clients could specify a list of desired responses that is less than
-  // or greater than the amount of requests they send on the stream. If this is the case
-  // and full duplex is specified then servers should behave as follows:
-  //
-  // If the total desired responses is less than the requests, the server should
-  // simply stop sending responses once the total desired is sent and continue to read
-  // messages until the client is finished sending.
-  //
-  // If the total desired responses is greater than the requests, the server should
-  // flush the remaining responses once the client is finished sending.
-  //
-  // Note that unequal desired responses and requests is irrelevant with half duplex calls.
   rpc BidiStream(stream BidiStreamRequest) returns (stream BidiStreamResponse);
 }
 
 // A definition of a response to be sent from a single-response endpoint.
 // Can be used to define a response for unary or client-streaming calls.
-message SingleResponseDefinition {
+message UnaryResponseDefinition {
   // Response headers to send
   repeated Header response_headers = 1;
 
@@ -127,7 +114,7 @@ message StreamResponseDefinition {
 
 message UnaryRequest {
   // The response definition which should be returned in the conformance payload
-  SingleResponseDefinition response_definition = 1;
+  UnaryResponseDefinition response_definition = 1;
 }
 
 message UnaryResponse {
@@ -149,7 +136,7 @@ message ClientStreamRequest {
   // Tells the server how to reply once all client messages are
   // complete. Required in the first message in the stream, but
   // should be ignored in subsequent messages.
-  SingleResponseDefinition response_definition = 1;
+  UnaryResponseDefinition response_definition = 1;
 
   // Additional data for subsequent messages in the stream.
   bytes extra = 2;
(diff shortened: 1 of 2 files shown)


proto/connectrpc/conformance/v1alpha1/service.proto
```diff
@@ -38,15 +45,21 @@ service ConformanceService {
   //
   // If the input stream is empty, the server's response will include no data,
   // only the request properties (headers, timeout).
-  rpc ClientStream(stream ClientStreamRequest) returns (ConformancePayload);
+  //
+  // Servers should only read the response definition from the first message in
+  // the stream and should ignore any definition set in subsequent messages.
```
smaye81, line 50: Should we allow clients to specify this in the subsequent messages and just ignore it? Or should we mandate that a server throws an error if anything is specified in subsequent messages?

proto/connectrpc/conformance/v1alpha1/service.proto
```diff
@@ -60,66 +73,116 @@ service ConformanceService {
...
+  // Note that clients could specify a list of desired responses that is less than
+  // or greater than the amount of requests they send on the stream. If this is the case
+  // and full duplex is specified then servers should behave as follows:
+  //
+  // If the total desired responses is less than the requests, the server should
+  // simply stop sending responses once the total desired is sent and continue to read
+  // messages until the client is finished sending.
+  //
+  // If the total desired responses is greater than the requests, the server should
+  // flush the remaining responses once the client is finished sending.
```
smaye81, line 86: This is up for debate probably. Should we allow a number of desired responses that is unequal to the requests sent by the client for full-duplex calls? It seems like we probably should, but I'm open to suggestions otherwise.

proto/connectrpc/conformance/v1alpha1/service.proto
```diff
@@ -60,66 +73,116 @@ service ConformanceService {
...
+  //
+  // Note that unequal desired responses and requests is irrelevant with half duplex calls.
+  rpc BidiStream(stream BidiStreamRequest) returns (stream BidiStreamResponse);
 }
 
-message UnaryRequest {
-  // response headers to send
+// A definition of a response to be sent from a single-response endpoint.
+// Can be used to define a response for unary or client-streaming calls.
+message SingleResponseDefinition {
```
smaye81, line 94: `Single...` meh. Open to other suggestions. Using `Unary` felt incorrect/misleading here since this could be used for client streams also.

proto/connectrpc/conformance/v1alpha1/service.proto
jhump: I think it's fine if the server just ignores the data. In practice, the actual test cases will never have requests that have that, so it's moot how the server behaves. If we want to trigger an "invalid argument" error in the server, that will be done from the response definition, not from sending a bad request.

proto/connectrpc/conformance/v1alpha1/service.proto
jhump: > // If the total desired responses is less than the requests, the server should
>  // simply stop sending responses once the total desired is sent and continue to read
>  // messages until the client is finished sending.

Might this cause the test to deadlock? If the client-under-test did the naive thing here and sends one request, waits for one response, etc, then after the last request, it will hang waiting on a response. I think this is a case where it's probably better to abort with an error if we get a message and have no corresponding response to send instead of risking a deadlock.

As far as what to do if number of responses > number of requests, looks good. In practice, the test cases will not have these kinds of anomalous configuration.

proto/connectrpc/conformance/v1alpha1/service.proto
jhump: I think `Unary` would be better. In my mind, a client stream does have a unary response (just not a unary request) 🤷

proto/connectrpc/conformance/v1alpha1/service.proto
```diff
@@ -131,19 +194,26 @@ message ConformancePayload {
...
     repeated Header request_headers = 1;
     // The timeout observed that was included in the request.
-    optional uint32 timeout_millis = 2;
+    optional uint32 timeout_ms = 2;
+    // The server should echo back all requests received.
+    // For unary and server-streaming requests, this should always contain a single request
+    // For client-streaming and half-duplex bidi-streaming, this should contain
+    // all client requests in the order received and be present in each response.
+    // For full-duplex bidirectional-streaming, this should contain all requests *so far* in the
+    // order received, meaning that this should grow as responses are sent.
```
jhump, line 203: > this should grow as responses are sent.

I would prefer this only be the requests observed since the last response. That way the test case data, in describing the expected responses, doesn't end up huge because earlier requests are repeated in every response.

jhump (approved):
I made some small recommendations, otherwise LGTM.

proto/connectrpc/conformance/v1alpha1/service.proto
smaye81: Makes sense. I'll leave the comments as-is.

proto/connectrpc/conformance/v1alpha1/service.proto
smaye81: Fair point. I just removed all these instructions from the comments since the test cases won't be constructed this way.

proto/connectrpc/conformance/v1alpha1/service.proto
smaye81: Cool. Updated the comment to reflect that.

PR #593, opened 2023-10-13
V2: Initial server implementation

smaye81:
This adds a first attempt at the reference implementation for a conformance server.

It adds the full implementation of all endpoints as well as an ability to run it via `cmd/server/main.go`.


internal/app/server/impl.go
```diff
@@ -0,0 +1,361 @@
...
+			for _, val := range header.Value {
+				stream.ResponseHeader().Add(header.Name, val)
+			}
+		}
+		// Set all requested response trailers on the response
+		for _, trailer := range responseDefinition.ResponseTrailers {
+			for _, val := range trailer.Value {
+				stream.ResponseTrailer().Add(trailer.Name, val)
+			}
+		}
```
smaye81, line 130: I don't like that this block of code is repeated a few times. There's just a bit of difficulty abstracting it with the APIs for `stream.ResponseTrailer()` and `resp.Trailer()`. We could definitely do it, but I'm not sure if it just makes it overly confusing vs. simply repeating the logic.

internal/app/server/server.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := h2Server.Shutdown(ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func newH1Server(h1Port string, handler http.Handler) *http.Server {
+	h1Server := &http.Server{
+		Addr:    ":" + h1Port,
```
jhump, line 114: I think this is equivalent to `"0.0.0.0:" + port`, which exposes the server on all network interfaces. But it is likely adequate to instead use `"127.0.0.1:" + port` to only expose the server on the loopback interface. It should probably at least _default_ to the loopback interface, and we could make it a parameter to bind to other interfaces if needed (unclear if it will be though).

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+
+// Sends a response T on the given stream. If delayMs is > 0, the function
+// will wait that many milliseconds before sending the message.
+func sendOnStream[T any](
+	ctx context.Context,
+	stream Stream[T],
+	resp *T,
+	delayMs uint32,
+) error {
+	var ticker *time.Ticker
```
jhump, line 269: You want a `Timer`, not a `Ticker`. A `Timer` is a one-shot timer. A `Ticker` is a recurring timer. Since this only cares about the first timer elapsing, no need for a ticker.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+		}
+	}
+
+	if err := stream.Send(resp); err != nil {
+		return fmt.Errorf("error sending on stream: %w", err)
+	}
```
jhump, line 284: The generic signature is only needed for this small bit. Maybe inline this into callers and instead name this `delay(context.Context, time.Duration)`?

Or maybe even simpler, remove this and just use `time.Sleep(duration)` at the call sites. Do we really need the delay to be able to wake up early if the context times out or is cancelled?

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+		switch rt := def.Response.(type) {
+		case *v1alpha1.UnaryResponseDefinition_Error:
+			return nil, createError(rt.Error)
+		case *v1alpha1.UnaryResponseDefinition_ResponseData, nil:
+			payload := initPayload(headers, reqs)
+
+			// If response data was provided, set that in the payload response
+			if rt, ok := rt.(*v1alpha1.UnaryResponseDefinition_ResponseData); ok {
+				payload.Data = rt.ResponseData
+			}
```
jhump, line 305: You are in the the case where you know this is the right type. And since you already have `rt := ` in the switch condition, the compiler already knows that `rt` is a `*v1alpha1.UnaryResponseDefinition_ResponseData`. So you shouldn't need this.
```suggestion
			payload.Data = rt.ResponseData
```

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+
+	if err := stream.Send(resp); err != nil {
+		return fmt.Errorf("error sending on stream: %w", err)
+	}
+	return nil
+}
+
+// Parses the given unary response definition and returns either
+// a built payload or a connect error based on the definition.
+func parseUnaryResponseDefinition(
```
jhump, line 290: In my mind, this isn't really "parsing". Maybe just call it `createResponse`?

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+		headerInfo = append(headerInfo, hdr)
+	}
+
+	// Set all observed request headers and requests in the response payload
+	payload.RequestInfo = &v1alpha1.ConformancePayload_RequestInfo{
+		RequestHeaders: headerInfo,
+		Requests:       reqs,
+	}
+	return payload
```
jhump, line 332: More idiomatic would be to do this entire thing as a single struct literal, instead of defining the var at the top and then mutating it at the bottom.
```suggestion
	headerInfo := make([]*v1alpha1.Header{}, 0, len(headers))
	for key, value := range headers {
		hdr := &v1alpha1.Header{
			Name:  key,
			Value: value,
		}
		headerInfo = append(headerInfo, hdr)
	}

	// Set all observed request headers and requests in the response payload
	return &v1alpha1.ConformancePayload{
		RequestInfo: &v1alpha1.ConformancePayload_RequestInfo{
			RequestHeaders: headerInfo,
			Requests:       reqs,
		}
	}
```

jhump: Also note that I used `make` to pre-size the slice, which is also idiomatic Go.

In general with slices, `slice := []sliceType{}` is _not_ idiomatic. Generally you'll either use a real slice literal, with the elements defined at construction time or use `make` to pre-size the slice's underlying buffer. If you don't know the size or the elements, then just leave it at the zero value to start with `var slice []sliceType`.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+// Sets all response headers and trailers onto the given response
+func setResponseHeadersAndTrailers[T any](def ConformanceRequest, resp *connect.Response[T]) {
```
jhump, line 336: I think we can simplify this -- get rid of the generics and make it work with unary and streaming responses -- by just making this a simple helper to translate `[]v1alpha1.Header` -> `http.Header`.
```suggestion
// Adds all header values in src to dest.
func addHeaders(
    src []*v1alpha1.Header,
    dest http.Header,
) {
```
The caller then does:
```go
addHeaders(respDefinition.ResponseHeaders, resp.Header())
addHeaders(respDefinition.ResponseTrailers, resp.Trailer())
```
And then you don't need the duplicated logic for the stream case.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+			return payload, nil
+		default:
+			return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("provided UnaryRequest.Response has an unexpected type %T", rt))
+		}
+	}
+	return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("no response definition provided"))
+}
+
+// Initializes a conformance payload
+func initPayload(headers http.Header, reqs []*anypb.Any) *v1alpha1.ConformancePayload {
```
jhump, line 315: Maybe instead `createPayload`. Or maybe even better: `createRequestInfo`, and it can just return `*v1alpha1.ConformancePayload_RequestInfo` instead of also returning the `*ConformancePayload` wrapper.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+)
+
+// ConformanceRequest is a general interface for all conformance requests (UnaryRequest, ServerStreamRequest, etc.)
+type ConformanceRequest interface {
+	GetResponseHeaders() []*v1alpha1.Header
+	GetResponseTrailers() []*v1alpha1.Header
+}
+
+// Stream represents a stream with the ability to set headers and trailers and send a message T
+type Stream[T any] interface {
```
jhump, line 39: I don't think these need to be exported. But I'm also not sure we need them at all. I think we can possibly tweak the signatures and responsibilities of the helper functions so that the helpers don't need to reach in and call these methods.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+
+		if err := sendOnStream[v1alpha1.ServerStreamResponse](
+			ctx,
+			stream,
+			resp,
+			responseDefinition.ResponseDelayMs,
+		); err != nil {
+			return err
+		}
+
```
jhump, line 155: I think you want to clear `payload.RequestInfo` here, so we only echo back the request in the first response.

I was reading back at the proto, and it's unclear what to do when `ResponseData` is empty. The comment says to send back no headers, trailers, or data in each response, but I guess I didn't read that closely enough during review to call that out -- "in each response" doesn't make sense, since we don't know how many to send back.

I definitely think it's useful to test a _zero_-sized response stream, which implies that if `ResponseData` is empty/unset, the server sends back no messages. That is what is actually implemented here, and I think that is the most sensible behavior. The one downside is that we have no way of echoing back the actual request data, to make sure it survives a round-trip. But I guess that is adequately tested in test cases where stream has at least one response message 🤷.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+			}
+			respNum++
+			reqs = nil
+		}
+	}
+
+	// If we still have responses left to send, flush them now. This accommodates
+	// both scenarios of half duplex (we haven't sent any responses yet) or full duplex
+	// where the requested responses are greater than the total requests.
+	if respNum < len(responseDefinition.ResponseData) {
```
jhump, line 230: No need for the `if` followed by a loop with the same condition. If you hoist the loop up, it will behave the same and the loop will run for exactly zero iterations.

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+		for i := respNum; i < len(responseDefinition.ResponseData); i++ {
```
jhump, line 231: No need to introduce a new counter var.
```suggestion
		for ; respNum < len(responseDefinition.ResponseData); respNum++ {
```

internal/app/server/impl.go
```diff
@@ -0,0 +1,374 @@
...
+	ctx context.Context,
+	req *connect.Request[v1alpha1.UnaryRequest],
+) (*connect.Response[v1alpha1.UnaryResponse], error) {
+	msgAsAny, err := asAny(req.Msg)
+	if err != nil {
+		return nil, err
+	}
+	payload, err := parseUnaryResponseDefinition(req.Msg.ResponseDefinition, req.Header(), []*anypb.Any{msgAsAny})
+	if err != nil {
+		return nil, err
```
jhump, line 57: You still need to populate response metadata. For unary RPCs, this is done by adding header metadata to a `*connect.Error`. It has a `Meta()` method that returns a mutable `http.Header`.

Sadly, the API provides no way to distinguish between headers and trailers in the error metadata. I think this is because, in practice, errors from unary RPCs often result in "trailers-only responses" in gRPC, which also cannot distinguish.

So I would do something like this here:
```suggestion
		addHeaders(req.Msg.ResponseDefinition.ResponseHeaders, err.Meta())
		addHeaders(req.Msg.ResponseDefinition.ResponseTrailers, err.Meta())
		return nil, err
```

cmd/server/main.go
```diff
@@ -16,222 +16,28 @@ package main
...
-		log.Fatalln(err)
-	}
-	if flags.h3Port != "" {
-		if err := h3Server.Close(); err != nil {
-			log.Fatalln(err)
-		}
-	}
-}
+	h1Port := flag.String("h1Port", "8080", "port for HTTP/1.1 traffic")
+	h2Port := flag.String("h2Port", "8081", "port for HTTP/2 traffic")
```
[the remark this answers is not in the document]
smaye81: Gotcha. My approach for this was to mainly just be able to get the server running and get the logic for each endpoint sound. I was thinking of expanding on how the server is actually fed the test cases in follow-up PRs if that's OK.

internal/app/server/server.go
```diff
@@ -0,0 +1,126 @@
...
+		// in response to preflight requests with the following configuration.
+		// It simply mirrors all headers listed in the `Access-Control-Request-Headers`
+		// preflight request header.
+		AllowedHeaders: []string{"*"},
+		// We explicitly set the exposed header names instead of using the wildcard *,
+		// because in requests with credentials, it is treated as the literal header
+		// name "*" without special semantics.
+		ExposedHeaders: []string{
+			"Grpc-Status", "Grpc-Message", "Grpc-Status-Details-Bin", "X-Grpc-Test-Echo-Initial",
+			"Trailer-X-Grpc-Test-Echo-Trailing-Bin", "Request-Protocol", "Get-Request"},
```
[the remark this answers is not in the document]
smaye81: Updated this to just use the wildcard.

internal/app/server/server.go
```diff
@@ -0,0 +1,126 @@
...
+	}).Handler(mux)
+
+	// Create servers
+	h1Server := newH1Server(h1Port, corsHandler)
+	h2Server := newH2Server(h2Port, mux)
+	done := make(chan os.Signal, 1)
+	errs := make(chan error, 2)
+	signal.Notify(done, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
```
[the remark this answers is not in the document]
smaye81: Yeah, this was from the previous server. I added a TODO about potentially revising this graceful shutdown.

internal/app/server/impl.go
smaye81: Yeah I made that overly complex. Removed in favor of a `time.Sleep`

internal/app/server/impl.go
smaye81: I think this is needed since the case is defined as `case *v1alpha1.UnaryResponseDefinition_ResponseData, nil:`. Otherwise, I get a compiler error: `rt.ResponseData undefined`

internal/app/server/impl.go
smaye81: I added this to unary and client stream requests. Should this also be included on the server and bidi requests in the event we return with an error?

internal/app/server/impl.go
```diff
@@ -0,0 +1,324 @@
...
+				Payload: &v1alpha1.ConformancePayload{
+					RequestInfo: requestInfo,
+					Data:        responseDefinition.ResponseData[respNum],
+				},
+			}
+			time.Sleep((time.Duration(responseDefinition.ResponseDelayMs) * time.Millisecond))
+
+			if err := stream.Send(resp); err != nil {
+				return connect.NewError(connect.CodeInternal, fmt.Errorf("error sending on stream: %w", err))
+			}
```
smaye81, line 211: This could probably be pulled out into its own function also. Lmk if you think it is worth it.

smaye81:
@jhump ready for round 2.

cmd/server/main.go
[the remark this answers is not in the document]
jhump: Absolutely. Sorry I wasn't clear -- I wasn't necessarily intending that to be a blocking comment. I'm totally good with adding rampant TODO comments. That way we can always grep for TODO close to the end to see if we are really done. So let's add a TODO for it, but fine to leave as is.

internal/app/server/impl.go
jhump: Nah. For those, the `stream` type the handler gets has ways to set headers and trailers separately. If you _do_ add metadata to the error, it will get merged into the trailers. But I don't think we want or need to do that here for those endpoints.

internal/app/server/impl.go
jhump: I guess I think or "parse" differently than that. It's fine either way -- not a strongly held opinion :)

jhump (approved):

smaye81:
OK, just going to leave the function name as-is for now. We can definitely rename it later if it's confusing.

PR #595, opened 2023-10-17
V2: Seed project with OSS skeleton

smaye81:
This adds the basic setup for OSS Go projects to the repo. Note the `cmd/` directory is removed because that is added in other PRs for implementing the server and client reference implementation.


jhump (approved):
Was it intentional to remove the files under `cmd`? I think that will cause a conflict with your other PR (#593). Also, seems fine to leave the client main as a placeholder until we get to it.

Also, for future PRs, let's prefix the PR name with "V2: ...", just so it's clear which PRs are ours and which impact the main branch until v2 is ready. (🤞 hopefully the PR title linter will accept that)

smaye81:
Yeah I did mean to remove those. Just because with the proto refactoring, they weren't compiling or passing lint, so I just removed them since the other branches are handling those. It's no biggie, I'll handle the merge conflicts when they arise.

PR #596, opened 2023-10-18
WIP: Read server config from stdin

smaye81:
This reads a `ServerCompatRequest` from stdin for standing up the test server. It then returns the server's status on stdout.

In addition, this adds a `--json` command line flag to allow for marshaling/unmarshaling of JSON format to make testing easier. Files to be used for testing are located in `testdata/`.

To run the server:

```go
// Runs a Connect server over HTTP/1.1
go run cmd/server/main.go --json < testdata/connecth1

// Runs a Connect server over HTTP/2
go run cmd/server/main.go --json < testdata/connecth2
```

Note right now, this just prints the host and port to terminal. Would like to get the internals working first and then being to refine the API.


internal/app/server/server.go
```diff
@@ -15,32 +15,68 @@
...
-	// TODO - The ports should not be given to the run call. Instead the server should read a
-	// ServerCompatRequest and start itself. It should then return the port numbers back on stdout
-	// as a ServerCompatResponse
-	h1Port := flag.String("h1Port", "8080", "port for HTTP/1.1 traffic")
-	h2Port := flag.String("h2Port", "8081", "port for HTTP/2 traffic")
+	rdr := bufio.NewReader(in)
+
+	var data string
+	// TODO - How should we read from 'in'? Using ReadString or ReadBytes?
+	// Assuming ReadBytes, but I used ReadString so i could test by redirecting a file
```
timostamm, line 48: Since we will typically be reading binary data, we should read bytes, not a string 🙂 

You can simply use `io.ReadAll` here if I'm not mistaken.

internal/app/server/server.go
```diff
@@ -66,55 +102,63 @@ func Run(ctx context.Context, args []string, in io.ReadCloser, out, err io.Write
...
+		err := server.Serve(ln)
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
 			errs <- err
 		}
 	}()
 
-	fmt.Printf("HTTP/1.1 server listening on port %s\nHTTP/2 server listening on port %s", *h1Port, *h2Port)
-
+	// TODO - Is there a race condition here where we reach the default case before any error is sent
+	// on the errs channel?
```
timostamm, line 129: It doesn't look quite right to me.

I'm not sure if it's idiomatic, but you can send `nil` on the channel on success and an error on failure. You wouldn't need select.

smaye81: The problem is that `server.Serve` blocks, so you can't send anything after. It's almost like I need a small delay before attempting to read from the `errs` channel to give the server time to start up and verify no errors occurred.

jhump: Errors at this stage are very rare. And, as you point out, there is nothing that can be done here since `Serve` blocks until the server shuts down if it is successful.

As currently implemented, the `default` case will almost certainly be taken: even if `Serve` returns an error, it is unlikely for the goroutine to have started, error returned, and error pushed into channel before we get to this point. So, since the `errs` channel is not ready, it will fall to the default case.

But another big problem here is that this immediately returns, which causes `Run` to return, which then cases `main` to return, which means that the server immediately stops and the process exits.

A better approach is to have this function return `(*http.Server, error)`. Then the caller can write the response to stdout on success and _then_ call `server.Serve`, but not in a goroutine, but blocking. That way, if `Serve` does unexpectedly return an error, it will cause `Run` to return an error, which should cause the program to exit with an error code, which I think is what would be expected.

smaye81:
OK, hopefully addressed some concerns. Let me know what you think @jhump @timostamm

internal/app/marshaler.go
```diff
@@ -0,0 +1,40 @@
+package app
+
+import (
+	"google.golang.org/protobuf/encoding/protojson"
+	proto "google.golang.org/protobuf/proto"
+)
+
+// Marshaler describes anything that can marshal and unmarshal proto messages.
+type Marshaler interface {
```
jhump, line 9: nit: no need to export this

Also, since it also unmarshals, this is typically named `codec`.

internal/app/marshaler.go
```diff
@@ -0,0 +1,40 @@
...
+
+// protoMarshaler marshals and unmarshals the Protobuf binary format.
+type protoMarshaler struct{}
+
+func (j *protoMarshaler) Marshal(msg proto.Message) ([]byte, error) {
+	return proto.Marshal(msg)
+}
+func (j *protoMarshaler) Unmarshal(b []byte, msg proto.Message) error {
+	return proto.Unmarshal(b, msg)
+}
```
jhump, line 40: I _think_ you can get away with no explicit methods if you embed types that provide them.
```suggestion
// jsonMarshaler marshals and unmarshals the JSON format.
type jsonMarshaler struct{
	protojson.MarshalOptions
	protojson.UnmarshalOptions
}

// protoMarshaler marshals and unmarshals the Protobuf binary format.
type protoMarshaler struct{
	proto.MarshalOptions
	proto.UnmarshalOptions
}
```

internal/app/marshaler.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// Marshaler describes anything that can marshal and unmarshal proto messages.
+type Marshaler interface {
+	Marshal(msg proto.Message) ([]byte, error)
+	Unmarshal(b []byte, msg proto.Message) error
+}
+
+// NewMarshaler returns a new marshaler.
+func NewMarshaler(json bool) Marshaler {
```
jhump, line 15: ditto: no need to export this

jhump (approved):

internal/app/marshaler.go
smaye81: Ah good call on both.

internal/app/marshaler.go
smaye81: This is actually in the `app` package and I was thinking the client would use it too, so I think we prob need it to stay exported. Let me know though if you'd rather it stay in the `server` package.

PR #605, opened 2023-10-25
Fix lint errors and wire up to CI

smaye81:
This performs some minor cleanup to the repo. Namely:

* Fixes any lint errors occurring when running `make lint`
* Adds an `Unimplemented` endpoint to `ConformanceService` so that unimplemented RPCs can be tested
* Cleans up the CI workflow, removing all the old Docker steps in favor of a slimmed down workflow that simply runs `test`, `lint`, and `checkgenerate`. Note, there are no tests right now, but some could be added in the future.
* Renames a few proto fields for consistency and/or clarity


smaye81:
@jhump if you'd rather I split these up into smaller PRs, that's cool too. Just figured since this was mostly inconsequential chore work, I'd combine them into one.

proto/connectrpc/conformance/v1alpha1/service.proto
```diff
@@ -88,6 +88,9 @@ service ConformanceService {
   // message that includes no data and only the request properties (headers,
   // timeout).
   rpc BidiStream(stream BidiStreamRequest) returns (stream BidiStreamResponse);
+  // A unary endpoint that the server should not implement and should instead
+  // return an unimplemented error when invoked.
+  rpc Unimplemented(UnimplementedRequest) returns (UnimplementedResponse);
```
jhump, line 93: I don't think this is sufficient. It is perhaps useful, but it only captures the case where the server handler _does_ provide the method: it's in the proto and thus in the generated server interfaces. So the only reason it will be unimplemented is because the server framework is supplying a default implementation.

We will also eventually need to test _completely unimplemented_ services and methods -- service and method names not known to the server at all. This can't be done right now using a simple `switch` in the reference client, so we can punt on it.

In the client implementation (not this PR), we probably want a TODO in the switch default that we'll eventually want to support other dispatch mechanisms -- that will allow us to do all kinds of things, like simulate exotic or intentionally incorrect client behavior, to verify how the server reacts.

Note that only the _reference_ will need to be able to craft requests for such unknown services and methods, since that is testing the _server_ behavior. Other clients-under-test would not need to worry about it.

.github/workflows/ci.yaml
```diff
@@ -1,9 +1,9 @@
 name: ci
 on:
   push:
-    branches: [main]
+    branches: [v2] # TODO: Switch to main once v2 is merged
   pull_request:
-    branches: [main]
+    branches: [v2] # TODO: Switch to main once v2 is merged
```
jhump, line 6: FWIW, I'm not sure this will actually take effect until it is merged to main.

So it might be a reasonable idea to propagate some of this config to a normal PR agains main branch, too. In that case, it should be `[main, v2]`.

proto/connectrpc/conformance/v1alpha1/client_compat.proto
```diff
@@ -45,11 +45,21 @@ message ClientCompatRequest {
...
+  // If true, it indicates the server should effectively interleave the
+  // stream so messages are sent in request->response pairs.
+  // If false, then the response stream will be sent once all request messages
+  // are finished sending with the only delays between messages
+  // being the optional fixed milliseconds defined in the response
+  // definition.
+  // This field is only relevant in the first message in the stream
+  // and should be ignored in subsequent messages.
+  // Note, this is only applicable to bidi endpoints.
+  bool full_duplex = 16;
```
jhump, line 62: Hmmm. I thought this change had already been made.

jhump (approved):

proto/connectrpc/conformance/v1alpha1/client_compat.proto
smaye81: That was in `server_compat.proto`. Changing it here to match.

proto/connectrpc/conformance/v1alpha1/service.proto
smaye81: Makes sense. Should I just remove this here? I feel like it's useful to test the scenario you mention: server knows about a service, but has not yet implemented it. But, I can certainly delete it if you don't think it provides that much value.

jhump: Seems fine to leave it.

timostamm: The old protos defined a unary and a server-streaming RPC that should not be implemented, and a separate service that should not be implemented, also with unary and server-streaming RPCs. It never seemed like a good solution to me.

I tend to think that it might be sufficient to:
1. have the reference client call unknown RPCs to verify they respond with either HTTP 404 or a gRPC / Connect unimplemented error
2. have the reference server respond with HTTP 404 and gRPC / Connect unimplemented error responses for certain test cases (based on test name in a request header)

This doesn't verify that servers reply with an in-protocol unimplemented error, but IMO this behavior could be considered an application-level or framework-level feature, not strictly a requirement by the protocol. On the upside, it doesn't introduce RPCs for the testee to reason about.

WDYT?

jhump: Yep, that was basically along the lines I was thinking for how to expand on this. Only the reference clients/servers need to handle these edge cases (i.e. a client-under-test will never need to be able to send a request for an unimplemented service or method).

PR #606, opened 2023-10-25
V2: Add reference client implementation

smaye81:
This adds a reference client implementation for the `ConformanceServiceClient`. The basic approach is that it reads a `ClientCompatRequest` from a reader and then configures a client based on that configuration. It then issues a request according to the `RequestMessages` in that config. After completion, it returns a `ClientCompatResponse`.

The implementation only supports unary and server-streaming requests for this initial approach to lessen the review burden. Follow-up PRs will implement client- and bidi-streaming.

In addition, a few files have been added to `testdata/` for running a test case via the client. Commands to see them in action:

First, start the server:

```bash
go run cmd/server/main.go --json < testdata/server/connecth1.json
```
The above will start a Connect-Go server serving HTTP/1.1 and will print the host and port to stdout. You can then use those values for the `host` and `port` values in any of the files in `testdata/client`. 

Once you have those updated, you can run:

```bash
# Run a happy path unary request (note, pipe to jq to see it formatted for JSON)
go run cmd/client/main.go --json < testdata/client/connecth1unary.json

# Run a unary request that returns an error
go run cmd/client/main.go --json < testdata/client/connecth1unaryerror.json

# Run a happy path server-streaming request that returns two responses with a 500ms delay between them
go run cmd/client/main.go --json < testdata/client/connecth1serverstream.json

# Run a server-streaming request that returns two responses with a 500ms delay between them and then an error
go run cmd/client/main.go --json < testdata/client/connecth1serverstreamerror.json
```


testdata/client/connecth1unaryerror
```diff
@@ -0,0 +1,59 @@
...
+              "baz"
+            ]
+          }
+        ],
+        "error": {
+          "code": 13,
+          "message": "unary failed",
+          "details": [
+            {
+              "@type": "type.googleapis.com/connectrpc.conformance.v1alpha1.Header",
```
smaye81, line 39: Used a `Header` type here just to test. I wanted to use a more real-world example such as `google.rpc.RetryInfo`, but I kept getting 'unable to resolve' errors and punted on it.

testdata/client/connecth1serverstream
jhump, line 1: Let's add `.json` extensions to all of these files.

testdata/client/connecth1unaryerror
jhump: Yep. Any real-world examples must refer to messages that your program actually knows about. To avoid those kinds of dependencies (and so that other client and server implementations that will be under test don't need to import/link in lots of other generated proto files/packages), I think it's best if we stick to messages that are defined in these conformance protos.

internal/app/client/client.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, errors.New("invalid url: %s" + urlString)
+	}
+
+	var transport http.RoundTripper
+	switch req.HttpVersion {
+	case v1alpha1.HTTPVersion_HTTP_VERSION_1:
+		transport = &http.Transport{}
+	case v1alpha1.HTTPVersion_HTTP_VERSION_2:
+		transport = &http2.Transport{
```
jhump, line 93: This seems like it will be a problem if/when we have 100s of test cases as this creates a new `http.Transport` for every request.

Maybe add a TODO to cache/re-use transports instead of creating a new one for every request?

internal/app/client/client.go
```diff
@@ -0,0 +1,144 @@
...
+		// Do nothing
+	case v1alpha1.Protocol_PROTOCOL_UNSPECIFIED:
+		return nil, errors.New("a protocol must be specified")
+	}
+
+	if req.Codec == v1alpha1.Codec_CODEC_JSON {
+		clientOptions = append(clientOptions, connect.WithProtoJSON())
+	}
+
+	// TODO - How do we configure each compression algo? i.e.
```
jhump, line 122: A TODO for this is fine.

But to answer the question: the reference client and server should support all of them, in case clients or servers under test support them. FWIW, the list of algorithms for the enum came from the actual protocol specs. Even though connect-go only supports "gzip" out-of-the-box, the Connect protocol spec also includes "br" and "zstd" (see _Content-Coding_ [here](https://connectrpc.com/docs/protocol/#unary-request)). And the gRPC protocol spec indicates "deflate" and "snappy" (see _Content-Coding_ [here](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests)).

In the future, there are implementations of all of these for Go:
* **br**: https://pkg.go.dev/github.com/andybalholm/brotli
* **zstd**: https://pkg.go.dev/github.com/klauspost/compress/zstd
* **deflate**: https://pkg.go.dev/compress/flate (in the Go stdlin actually)
* **snappy**: https://pkg.go.dev/github.com/golang/snappy

And they all have fairly similar APIs, similar to the decompressor/compressor reader/writer APIs of the gzip package (which is what connect-go includes by default).

internal/app/client/client.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"connectrpc.com/conformance/internal/app"
+	"connectrpc.com/conformance/internal/gen/proto/connect/connectrpc/conformance/v1alpha1/conformancev1alpha1connect"
+	v1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
+	"connectrpc.com/connect"
+	"golang.org/x/net/http2"
+)
+
+// Wrapper defines a wrapper around a client which can invoke requests based on a ClientCompatRequest.
+type Wrapper interface {
```
jhump, line 36: This doesn't need to be exported. In fact, I don't think you need an interface at all since there's only one implementation (`conformanceClientWrapper`) -- so callers could just refer directly to that one concrete type.

internal/app/client/conformance_client.go
```diff
@@ -0,0 +1,191 @@
...
+	"net/http"
+	"net/url"
+
+	"connectrpc.com/conformance/internal/app"
+	"connectrpc.com/conformance/internal/gen/proto/connect/connectrpc/conformance/v1alpha1/conformancev1alpha1connect"
+	v1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
+	"connectrpc.com/connect"
+)
+
+type conformanceClientWrapper struct {
```
jhump, line 29: In addition to removing the `Wrapper` interface, I think we can also give this a shorter name. How about `invoker`?

internal/app/client/conformance_client.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		resp, err := w.serverStream(ctx, req)
+		if err != nil {
+			return nil, err
+		}
+		return resp, nil
+	default:
+		// TODO: Should this be a returned 'error' or via the ClientCompatResponse?
+		// We should probably treat 'error' here as something independent of invoking a request
+		// So any internal error that doesn't involve actually calling an RPC. In that case, then,
```
jhump, line 59: I think ideally this method would just return `(*ClientResponseResult, error)`. So the known cases above, would always return `*ClientResponseResult` when they were able to actually invoke the RPC, and wrap any resulting error in the result. Other kinds of errors it can return.

The caller then wraps the return value into `*ClientCompatResponse`. So if an error is returned, the `err.Error()` message gets automatically stored into `ClientCompatResponse.error.message` and then the response is written to stdout by the caller.

BTW, mind adding a TODO for client-stream and bidi, too?

internal/app/client/conformance_client.go
```diff
@@ -0,0 +1,191 @@
...
+			return nil, err
+		}
+		return resp, nil
+	case "ServerStream":
+		if len(req.RequestMessages) != 1 {
+			return nil, errors.New("server streaming calls must specify exactly one request message")
+		}
+		resp, err := w.serverStream(ctx, req)
+		if err != nil {
+			return nil, err
```
jhump, line 53: Per comment added in default case below: since we've actually invoked the method at this point, we should try to wrap `err` into `ClientCompatResponse.response.error`.

internal/app/client/conformance_client.go
```diff
@@ -0,0 +1,191 @@
...
+
+	request := connect.NewRequest(ssr)
+
+	// Add the specified request headers to the request
+	app.AddHeaders(req.RequestHeaders, request.Header())
+
+	stream, err := w.client.ServerStream(ctx, request)
+	// TODO - should this error be added to the clientcompatresponse or returned here?
+	// IMO, the error returned from this function represents an internal error independent
+	// of anything invoking the service but it's unclear what this err is vs. a stream.Err()
```
jhump, line 146: See comments above, which hopefully clarify this. I think this should do what it's doing and return the error. And in `Invoke`, it gets stored into a the `error` field of a `*ClientResponseResult`.

internal/app/errors.go
```diff
@@ -0,0 +1,92 @@
...
+
+	v1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
+	"connectrpc.com/connect"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+// ConvertErrorToConnectError converts the given error to a Connect error
+// If err is nil, function will also return nil. If err is not
+// of type connect.Error, a Connect error of code Unknown is returned.
+func ConvertErrorToConnectError(err error) *connect.Error {
```
jhump, line 28: nit: none of these to be exported

testdata/client/connecth1unaryerror
smaye81: Gotcha. I actually defined a `RetryInfo` in our protos too and got that same message, but I assume I was just doing something wrong.

internal/app/client/client.go
smaye81: Gotcha. Thank you. Reworded the TODO and will add support in a follow-up

internal/app/client/client.go
smaye81: Updated. I figured maybe this would be helpful if we ever need additional implementations?

internal/app/client/conformance_client.go
smaye81: Yeah I like that approach actually. Thanks. I updated the code and added some clarifying comments to the various error handling and what the difference is. Hopefully it clears things up (at least it does for me). 

Also added the requested TODOs.

internal/app/errors.go
smaye81: These are actually in the `app` package since the functionality is needed for both server and client. I can move elsewhere if you like though.

internal/app/client/conformance_client.go
```diff
@@ -0,0 +1,173 @@
...
+	"net/http"
+	"net/url"
+
+	"connectrpc.com/conformance/internal/app"
+	"connectrpc.com/conformance/internal/gen/proto/connect/connectrpc/conformance/v1alpha1/conformancev1alpha1connect"
+	v1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
+	"connectrpc.com/connect"
+)
+
+type Invoker struct {
```
smaye81, line 29: Lint requested this to be exported since `NewInvoker` was returning an unexported type. However, this is only used within the `client` package, so we could ignore that rule for this case. Either way works for me.

jhump: The solution is to unexport `NewInvoker`, too 😛 

In general, only export what is actually used/needed outside of the package.

testdata/client/connecth1unaryerror
jhump: If you defined it in a different package, that would be the issue: the Go runtime is only to do JSON decoding of `Any` messages for messages defined in packages that are actually _imported_ by the program. So if you added new messages in a different folder, but then never updated any of this code to _import_ that package, the types will not be available.

FWIW, this is only an issue when decoding JSON. For the protobuf binary format, `Any` messages can be used and de-serialized without knowing the schema of the message therein. So we should actually test this in a test case, that we the server can reply correctly even if an error details contains an unknown message. To do that, the test runner itself will need to know about some protos that the clients/servers under test (including the reference client and server) do _not_ know about, so that it can actually decode them (from YAML configs).

But I think we can worry about that later. We could even have the test runner _dynamically_ load (or even generate) some protos, so it's _guaranteed_ that the client or server under test _can't_ know about them, and then verify that it can return the errors correctly anyhow.

Anyhow, I say all of this because the actual gRPC/gRPC-Web/Connect format for errors does not require the server or client knowing the schema, since they use the binary format.

internal/app/client/client.go
jhump: We can always change type signatures to use an interface in the future when we get there. Though, TBH, I can't think of anything that would necessitate a second implementation off the top of my head.

And since none of this importable library code, we don't have to be concerned about changing signatures and breaking users. So we can freely change signatures and types in the future if need be.

internal/app/errors.go
jhump: Ah, my bad.

internal/app/client/conformance_client.go
smaye81: Ugh, duh.  Yeah, sorry about that. Fixed.

internal/app/client/client.go
```diff
@@ -0,0 +1,164 @@
...
+func Run(ctx context.Context, _ []string, inReader io.ReadCloser, outWriter io.WriteCloser) error {
+	json := flag.Bool("json", false, "whether to use the JSON format for marshaling / unmarshaling messages")
+
+	flag.Parse()
+
+	// Read the server config from  the in reader
+	data, err := io.ReadAll(inReader)
+	if err != nil {
+		return err
+	}
```
jhump, line 48: Sorry I didn't catch this before. And double-sorry since it was not spelled out anywhere in any of the comments (though I think may have been alluded to in a comment in the design in Notion).

But the intent here was that a client might read _many_ RPC definitions. We don't want the overhead of spawning a new process for every single request.

The format was to be varint-length-prefixed protos.

So the client should be reading input in a loop, something like so:
```go
// pseudo-code-ish
for {
    // read varint-encoded prefix that indicates message length
    length, err := readVarintLength(inReader)
    if errors.Is(err, io.EOF) { return nil /* all done */ }
    if err != nil { return err }
    // Read the message
    data := make([]byte, length)
    if _, err := io.ReadFull(inReader, data); err != nil {
        return err
    }

    // ... Now we can unmarshal the request, issue the RPC,
    //     write the response results, and then go to top of
    //     loop to read another request ...
}
```

This was the side discussion where Timo asked about the format here: https://www.notion.so/Connect-Conformance-Tests-Design-0a279279a89c48559453ac3c7474b063?d=3e6bb2676deb43389dfa0546bb760ac6&pvs=4#4f3088105113406cbf8c371dda1ac999

And now that I think about, "varint-encoded" prefix is unnecessarily complicated. Instead, it could just read four bytes as network-byte-order `uint32` for the size, and then read that many more bytes for the actual message. I say varint is too complicated because it looks like there isn't actually any decent implementation to do a streaming read of a varint-encoded number from an `io.Reader`. So maybe it would just be easier to make it a fixed-length integer...

FWIW, I'm not sure how this would work with JSON inputs. Maybe we can keep this as is for now and just add a TODO, and then we can follow up and make sure that whatever format this reads is the same as the format that the test runner actually writes.

(Aside: it is only necessary for the client to act this way. A server will only ever read/write a single protobuf message, not a stream of them, so we should leave it as is.)

internal/app/client/client.go
```diff
@@ -119,26 +140,25 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
...
+	// TODO - Add support for other compression algos
 	switch req.Compression {
 	case v1alpha1.Compression_COMPRESSION_GZIP:
 		clientOptions = append(clientOptions, connect.WithSendGzip())
-	case v1alpha1.Compression_COMPRESSION_IDENTITY:
 	case v1alpha1.Compression_COMPRESSION_BR:
 	case v1alpha1.Compression_COMPRESSION_ZSTD:
 	case v1alpha1.Compression_COMPRESSION_DEFLATE:
 	case v1alpha1.Compression_COMPRESSION_SNAPPY:
+	case v1alpha1.Compression_COMPRESSION_IDENTITY:
```
jhump: We should be returning an error for the unsupported compression algorithms instead of doing nothing.

Also, these all happen to do nothing, but not because they fall-through. There is no fall-through in Go switch statements unless you do so explicitly with the `fallthrough` keyword.

The preferred way to instead have multiple values map to the same case is to use comma-separate list of case values:
```suggestion
	case v1alpha1.Compression_COMPRESSION_BR, v1alpha1.Compression_COMPRESSION_ZSTD.
		v1alpha1.Compression_COMPRESSION_DEFLATE, v1alpha1.Compression_COMPRESSION_SNAPPY:
		// return an error since we don't actually support any of these yet
	case v1alpha1.Compression_COMPRESSION_IDENTITY, v1alpha1.Compression_COMPRESSION_UNSPECIFIED:
		// do nothing
```

internal/app/client/client.go
```diff
@@ -119,26 +140,25 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
...
 	case v1alpha1.Compression_COMPRESSION_UNSPECIFIED:
 		// Do nothing
 	}
 
-	var wrapper Wrapper
 	switch req.Service {
 	case conformancev1alpha1connect.ConformanceServiceName:
-		wrapper = NewConformanceClientWrapper(transport, serverURL, clientOptions)
+		wrapper := NewInvoker(transport, serverURL, clientOptions)
+		return wrapper.Invoke(ctx, req)
```
jhump: small nit -- var is still named "wrapper" which no longer really makes sense. Also, var isn't needed
```suggestion
		return NewInvoker(transport, serverURL, clientOptions).Invoke(ctx, req)
```

testdata/client/connecth1unary.json
```diff
@@ -4,7 +4,7 @@
   "protocol": "PROTOCOL_CONNECT",
   "codec": "CODEC_JSON",
   "host": "127.0.0.1",
-  "port": 51356,
+  "port": 53289,
```
jhump, line 7: You might want to add a `--port` arg to server, that defaults to zero (to pick ephemeral port). That way you can start it up with a stable port, so these files are easier to work with.

jhump (approved):

testdata/client/connecth1unary.json
smaye81: Yeah good call.  Added.

internal/app/client/client.go
smaye81: Sounds good. I added a TODO and we can address it when the runner is ready for primetime.

PR #607, opened 2023-10-26
V2: Test harness

jhump:
This adds a `connectconformance` command that handles running a reference client or server and spawning a client or server under test via given args.

This is a draft PR because this hasn't been tested yet -- it passes `make test lint`, but there are no actual tests yet, so it without-doubt does not actually work 😂 

This has a decent bit of logic in it to interpret the `Config` message (defined in `config.proto`) and then also interpret `TestSuite` messages (defined `suite.proto`), explode them across all applicable permutations (HTTP version, protocol, codec, compression, etc), and then filter them according to the given config.

There are plenty of TODOs in here as well. In particular, I'd like to re-architect this at some point so it can spin up multiple servers in parallel and potentially run the tests much more quickly if the client also supports issuing RPCs in parallel.


jhump:
cc @smaye81

jhump:
I'm going to start splitting this up into smaller PRs and also adding some test cases to places where there is logic (expanding a config into permutations of "config cases" and then also expanding test suites into permutations of test cases filtered by currently applicable config cases).

So closing this now. This will likely turn into 3 or 4 smaller branches.

PR #608, opened 2023-10-26
Implement ClientStream RPC

smaye81:
This adds support for the `ClientStream` RPC to the conformance client implementation.


internal/app/client/conformance_client.go
```diff
@@ -54,8 +55,14 @@ func (i *invoker) Invoke(
 		}
 		return resp, nil
 	case "ClientStream":
-		// TODO - Implement ClientStream
-		return nil, errors.New("client streaming is not yet supported")
+		if len(req.RequestMessages) < 1 {
+			return nil, errors.New("client streaming calls must specify at least one request message")
+		}
```
jhump, line 60: I don't think we want this check. Instead we should settle on (and document) what the server should do when it gets a client stream or bidi stream with zero request messages. That is actually a valid RPC case that we should be able to test.

I would propose that in both cases the server send back a single response where `payload.data` is empty.

And additional related behavior -- what to do when the response definition for server or bidi stream indicate zero responses: the server should serialize the `ConformancePayload` to bytes and store it as a binary response header, like `conformance-payload-bin`. We'll need some way to model this in the test result expectations of a test case, to tell the test harness that it should expect the payload in a header instead of a message.

internal/app/client/conformance_client.go
```diff
@@ -160,6 +167,55 @@ func (i *invoker) serverStream(
...
+
+	// Add the specified request headers to the request
+	app.AddHeaders(req.RequestHeaders, stream.RequestHeader())
+
+	for _, msg := range req.RequestMessages {
+		csr := &v1alpha1.ClientStreamRequest{}
+		if err := msg.UnmarshalTo(csr); err != nil {
+			return nil, err
+		}
+		if err := stream.Send(csr); err != nil && errors.Is(err, io.EOF) {
```
jhump, line 184: I think this should sleep for the delay indicated in `req` before sending, right?

jhump (approved):
👍

internal/app/client/conformance_client.go
smaye81: Makes sense. I removed that check

internal/app/client/conformance_client.go
smaye81: Ah yeah, good call. I actually updated the proto for `ClientCompatRequest` also. I previously had a TODO that the delay field there could be removed, but now I realize that should specify the delay for _requests_.

PR #609, opened 2023-10-27
WIP: V2: Implement BidiStream RPC

smaye81:
This is an initial stab at the `BidiStream` client call. Currently, a work in progress while error handling is ironed out.


internal/app/client/conformance_client.go
```diff
@@ -218,6 +221,118 @@ func (i *invoker) clientStream(
...
+	// If protoErr is nil, then the send loop terminated successfully
+	if protoErr == nil {
+		// Close the send side of the stream
+		if err := stream.CloseRequest(); err != nil {
+			protoErr = app.ConvertErrorToProtoError(err)
+		}
+	}
+
+	// If there were no errors encountered yet, then receive any additional responses
+	if protoErr == nil {
```
smaye81, line 295: Feels gross to check `protoErr` here again, but this is basically checking "the send loop completed successfully, but closing the send side of the stream failed, so don't try to receive any additional responses"

smaye81:
@jhump this is the first stab at Bidi. I am positive this needs refining / refactoring, but I wanted to get something up so we could iterate on it. I left a few TODOs also where I was unsure about certain things.

internal/app/client/conformance_client.go
```diff
@@ -218,6 +221,118 @@ func (i *invoker) clientStream(
...
+			return nil, err
+		}
+		if firstSend {
+			fullDuplex = bsr.FullDuplex
+		}
+		if err := stream.Send(bsr); err != nil && errors.Is(err, io.EOF) {
+			// TODO - Is this correct? Docs say the returned error will wrap EOF
+			// and clients should call Receive to get the error so assuming this error
+			// is the unwrapped error from the send call?
+			// Call receive to get the error
```
jhump, line 258: Yep, this looks right. The only thing I'd suggest, just to be double-extra-safe, is to use the send error just in case the call to `Receive` returns nil (it shouldn't... but just in case).

jhump: BTW, this is an interesting case that isn't adequately covered in `ClientResponseResult`: we should probably add a field to that message to indicate the number of messages that were in the test case request definition but that _couldn't be sent_, due to an error occurring before finishing the upload. (Technically, we'd want that number for client stream operations, too.)

internal/app/client/conformance_client.go
jhump: Instead of `if protoErr == nil`, I think it would be better to invert the condition and early-return. So `if protoErr != nil`, we're done.

To do that w/out lots of boiler-plate, you could use a deferred function that does all of the stuff at the end that is needed using named returns.

So if the signature used named returns like so:
```go
func (i *invoker) bidiStream(
	ctx context.Context,
	req *v1alpha1.ClientCompatRequest,
) (result *v1alpha1.ClientResponseResult, err error) {
	result = &v1alpha1.ClientResponseResult{}
```
Then you can directly mutate the result as you collect payloads, etc (e.g. use `result.Payloads = append(result.Payloads, msg.Payload)`).

So early return could be something like this:
```go
if protoErr != nil {
	result.Error = protoErr
	return result, nil
}
```

Finally, to catch the headers and trailers at the end, you could use a defer function installed at the beginning:
```go
defer func() {
	if result != nil {
		result.Headers = app.ConvertToProtoHeader(stream.ResponseHeader())
		result.Trailers = app.ConvertToProtoHeader(stream.ResponseTrailer())
	}
}()
```

internal/app/client/conformance_client.go
smaye81: Added in the extra check and also added a TODO to the proto for `ClientResponseResult` that we may want that field for unsent messages.

internal/app/client/conformance_client.go
```diff
@@ -221,23 +221,29 @@ func (i *invoker) clientStream(
 	}, nil
 }
 
-func (i *invoker) bidiStream(
+func (i *invoker) bidiStream( //nolint:nonamedreturns
```
jhump, line 224: IMO, the "nonamedreturns" linter is _never_ valuable. Let's just add it to the exclude list, like so:
https://github.com/bufbuild/httplb/blob/main/.golangci.yml#L48

internal/app/client/conformance_client.go
```diff
@@ -246,91 +252,88 @@ func (i *invoker) bidiStream(
...
+	if protoErr != nil {
+		result.Error = protoErr
+		return result, nil
 	}
 
-	// Read headers and trailers from the stream
-	headers = app.ConvertToProtoHeader(stream.ResponseHeader())
-	trailers = app.ConvertToProtoHeader(stream.ResponseTrailer())
+	// TODO - Does this unwrap the error from the receive call or is this
+	// a separate error that needs handled?
```
jhump, line 330: You always return any error from a receive call, so if this returns an error, it will be something new I think. Unclear though since a non-nil error from receive effectively means "end of stream", so there may be nothing else that can happen here.

In any event, I guess I'm saying that the code you have here looks good to me.

internal/app/client/conformance_client.go
```diff
@@ -246,91 +252,88 @@ func (i *invoker) bidiStream(
...
+	if err := stream.CloseResponse(); err != nil {
+		result.Error = app.ConvertErrorToProtoError(err)
+	}
 
 	// TODO - How can we distinguish whether we properly processed full vs. half duplex?
```
jhump, line 335: I'm not sure what this is asking. If the above works without returning an error, I'm pretty sure it's been processed correctly.

jhump (approved):

PR #617, opened 2023-10-30
V2: Update protos and add comments so config and test suite definitions are more clear

jhump:
This also streamlines a few other things, like flattening the `ServerCompatResponse` and adding a few new flags to the config and test suite messages.

You can largely ignore the diff stats: this isn't really that big of a PR, but since it changes protos, it also changes generated code which is what leads to the bloated stats.


proto/connectrpc/conformance/v1alpha1/client_compat.proto
```diff
@@ -40,30 +40,20 @@ message ClientCompatRequest {
...
-  // stream so messages are sent in request->response pairs.
-  // If false, then the response stream will be sent once all request messages
-  // are finished sending with the only delays between messages
-  // being the optional fixed milliseconds defined in the response
-  // definition.
-  // This field is only relevant in the first message in the stream
-  // and should be ignored in subsequent messages.
-  // Note, this is only applicable to bidi endpoints.
-  // TODO - This could potentially be removed bc this is in the BidiStreamRequest
-  // which would be in the request_messages field
```
jhump, line 65: This is now implied by the stream type field. I suppose I might preserve the TODO that this is technically still in two places -- in this message and in the response definition of the request.

proto/connectrpc/conformance/v1alpha1/client_compat.proto
```diff
@@ -80,16 +70,20 @@ message ClientResponseResult {
...
   Error error = 3;
-  // In case the client cannot decode Any from JSON, it should instead return the received JSON
-  repeated google.protobuf.Struct error_details_raw = 7;
   repeated Header response_trailers = 4;
+  // When processing an error from a Connect server, this should contain
+  // the actual JSON received on the wire. Clients under test do not need
+  // to populate this field. It is primarily used by the reference client,
+  // so that a server-under-test's on-the-wire error can be examined and
+  // validated.
+  google.protobuf.Struct connect_error_raw = 5;
```
jhump, line 81: I think this name is a little more descriptive, since this field only applies to the Connect protocol.

proto/connectrpc/conformance/v1alpha1/config.proto
```diff
@@ -89,5 +113,6 @@ enum StreamType {
   STREAM_TYPE_UNARY = 1;
   STREAM_TYPE_CLIENT_STREAM = 2;
   STREAM_TYPE_SERVER_STREAM = 3;
-  STREAM_TYPE_BIDI_STREAM = 4;
+  STREAM_TYPE_HALF_DUPLEX_BIDI_STREAM = 4;
+  STREAM_TYPE_FULL_DUPLEX_BIDI_STREAM = 5;
```
jhump, line 117: Since it's possible for an implementation to choose to support half-duplex but not full-duplex (like if using HTTP/1.1), this seems like a reasonable split to make.

proto/connectrpc/conformance/v1alpha1/server_compat.proto
```diff
@@ -42,15 +42,8 @@ message ServerCompatRequest {
 
 // The outcome of one ServerCompatRequest.
 message ServerCompatResponse {
-  oneof result {
```
jhump, line 45: I think we talked about this change. I went ahead and made it since it simplifies the test runner to now have to consider the error case this way.

proto/connectrpc/conformance/v1alpha1/server_compat.proto
```diff
@@ -42,15 +42,8 @@ message ServerCompatRequest {
...
-    ServerListeningResult listening = 1;
-    ServerErrorResult error = 2;
-  }
-}
-
-message ServerListeningResult {
   string host = 1;
-  string port = 2;
+  uint32 port = 2;
```
jhump, line 46: Oops, I think this may have been copy+pasta. It should have been a number.

proto/connectrpc/conformance/v1alpha1/config.proto
```diff
@@ -19,35 +19,59 @@ package connectrpc.conformance.v1alpha1;
...
-  // If non-empty, the set of stream types that are supported. This is
-  // usually for clients, since some client environments may not be
-  // able to support certain kinds of streaming operations, especially
+  // If empty, "gzip" alone is assumed.
+  repeated Compression compressions = 4;
+  // If absent, true is assumed.
+  optional bool supports_h2c = 5;
+  // If absent, true is assumed.
+  optional bool supports_tls = 6;
+  // If absent, true is assumed. If false, implies that gRPC protocol is not allowed.
```
smaye81, line 48: Maybe we could eventually use protovalidate here to validate things like this? i.e. if this is false, `protocols` cannot contain `gRPC`? Or maybe we'd just display a runtime error message 🤷

proto/connectrpc/conformance/v1alpha1/config.proto
```diff
@@ -19,35 +19,59 @@ package connectrpc.conformance.v1alpha1;
...
-  repeated Compression compression_algorithms = 4;
-  bool supports_h2c = 5;
-  bool supports_tls = 6;
-  bool supports_trailers = 7;
-  bool supports_connect_get = 8;
-  bool requires_connect_version_header = 9;
-  // If non-empty, the set of stream types that are supported. This is
-  // usually for clients, since some client environments may not be
-  // able to support certain kinds of streaming operations, especially
+  // If empty, "gzip" alone is assumed.
```
smaye81, line 42: I assume this is because a feature implies all possible compression algos that a client/server could support and since gzip is available by default, it makes sense to assume that. Just wanted to confirm since for the `ClientCompatRequest`, an unspecified compression implies `identity`.

proto/connectrpc/conformance/v1alpha1/client_compat.proto
smaye81: Yeah I'd agree with that. I can see having the two fields, but we could at least have a comment for how they interact and why they both exist.

smaye81 (approved):

proto/connectrpc/conformance/v1alpha1/client_compat.proto
jhump: We could even make the code that loads and processes test suite configs auto-populate the "full_duplex" field in the test case response definition based on this stream type, which would eliminate the redundancy.

proto/connectrpc/conformance/v1alpha1/config.proto
jhump: FWIW, I currently implement those sorts of rules by hand in the implementation for processing config (in #618).

Many of the rules enforced _could_ in fact be validated using protovalidate. But some of them are harder -- like interactions between supported features and explicitly included cases, where an invalid combination occurs. So for now, it's all in code 🤷

I'll add a TODO.

proto/connectrpc/conformance/v1alpha1/config.proto
jhump: Yep, my bad. I'll revise this to indicate that if empty, both "identifty" and "gzip" are assumed.

PR #619, opened 2023-10-31
V2: Load test cases from embedded test suite data

jhump:
This PR adds the logic for embedding test suite data and for loading the data, filtering it according to active config cases, and then grouping by server process. In this case, a "server process" identifies a `ServerCompatRequest` and thus a single server process that will be started by the test runner.

### DIFF
--- a/internal/app/connectconformance/test_case_library_test.go
+++ b/internal/app/connectconformance/test_case_library_test.go
@@ -1,0 +1,292 @@
+// Copyright 2023 The Connect Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connectconformance
+
+import (
+	"sort"
+	"testing"
+
+	"connectrpc.com/conformance/internal/app/connectconformance/testsuites"
+	conformancev1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadTestCases(t *testing.T) {
+	t.Parallel()
+
+	testData := map[string]string{
+		"basic.yaml": `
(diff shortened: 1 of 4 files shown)


internal/app/connectconformance/test_case_library_test.go
```diff
@@ -0,0 +1,292 @@
...
+					"Basic/HTTPVersion:1/Protocol:PROTOCOL_CONNECT/Codec:CODEC_PROTO/Compression:COMPRESSION_IDENTITY/basic-unary",
+					"Connect GET/HTTPVersion:1/Codec:CODEC_PROTO/Compression:COMPRESSION_IDENTITY/connect-get-unary",
+					"Connect Version Required (client)/HTTPVersion:1/Codec:CODEC_PROTO/Compression:COMPRESSION_IDENTITY/unary-without-connect-version-header",
+				},
+				{
+					protocol:    conformancev1alpha1.Protocol_PROTOCOL_CONNECT,
+					httpVersion: conformancev1alpha1.HTTPVersion_HTTP_VERSION_1,
+					useTLS:      true,
+				}: {
+					"TLS/HTTPVersion:1/Protocol:PROTOCOL_CONNECT/Codec:CODEC_PROTO/Compression:COMPRESSION_IDENTITY/tls-unary",
```
jhump, line 181: Open question here: in this test, we can see that it will only run the TLS-**specific** test case against the TLS server. My thinking is that we'd define just basic "success" RPCs over TLS as sufficient evidence that TLS is functioning.

But another interpretation might be to run _all_ test cases through both plaintext and TLS, when TLS is supported. Overkill? WDYT?

smaye81: My opinion would be to leave it out for now. The ability to serve plaintext over a secure server seems like more of an outlier/specialty case so it feels like overkill. We could always add them in later if we think it's worthwhile to test. Just my 2 cents.

internal/app/connectconformance/testsuites/testsuites.go
```diff
@@ -0,0 +1,56 @@
...
+	err := fs.WalkDir(testSuiteFS, ".", func(currentPath string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if entry.IsDir() || strings.ToLower(path.Ext(entry.Name())) != ".yaml" {
+			return nil
+		}
+		data, err := testSuiteFS.ReadFile(currentPath)
+		if err != nil {
+			return fmt.Errorf("failed to load test suite date file %s: %w", currentPath, err)
```
smaye81, line 47: nit: did you mean `data`?

smaye81 (approved):

internal/app/connectconformance/testsuites/testsuites.go
jhump: Yep, I did. Fixed.

PR #620, opened 2023-10-31
Remove npm and docker from dependabot

smaye81:
This updates the dependabot yaml to remove NPM and Docker.


timostamm (approved):
:doublestamp:

PR #621, opened 2023-10-31
V2: Add compression support for servers and clients

smaye81:
This begins to add compression support (other than gzip) for servers and clients. The first algorithm added is `zstd`. Once the initial setup is ironed out, the others can be added easily.


internal/compression/compression.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compression
+
+// The IANA names for supported compression algorithms
```
smaye81, line 17: I was thinking we could also have caches here for compressors/decompressors so that we aren't creating them over and over again for each test case. Not sure if that's overkill or not.

internal/compression/zstd.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// ZstdCompressor is a thin wrapper around a zstd Encoder.
+type ZstdCompressor struct {
+	*zstd.Encoder
+}
+
+// NewZstdCompressor returns a new Zstd Compressor.
+func NewZstdCompressor() connect.Compressor {
+	w, _ := zstd.NewWriter(nil)
```
smaye81, line 55: We have to swallow this error here because the `connect.WithAcceptCompression` method expects a function that just returns a `connect.Compressor` (not a compressor and error) and the same goes for the decompressor.

internal/compression/zstd.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (c *ZstdDecompressor) Read(bytes []byte) (int, error) {
+	return c.decoder.Read(bytes)
+}
+func (c *ZstdDecompressor) Reset(rdr io.Reader) error {
+	return c.decoder.Reset(rdr)
+}
+func (c *ZstdDecompressor) Close() error {
+	c.decoder.Close()
+	return nil
```
smaye81, line 37: Always returning `nil` here because the `zstd` function for `Close` doesn't return an error.

internal/compression/zstd.go
jhump: This is bound to cause a panic if an error ever does occur. Eating errors is almost never acceptable -- often better to panic so that at least the code fails fast with a good error message instead of possibly panic'ing later with an inexplicable nil-pointer dereference.

In this case, instead of panic'ing or ignoring, there is another option (which should also be used above in `NewZstdDecompressor`): return a sentinel, like an `errorCompressor` or `errorDecompressor`. Those types can have a single `err error` field and _always_ return that error from every method. That way you can see the error that happened here, but it will be delayed to the point of first actual use of the compressor/decompressor.

smaye81: Awesome idea, thank you. Updated.

internal/compression/sentinel.go
```diff
@@ -0,0 +1,51 @@
...
+// which will return an error upon first use.
+type errorCompressor struct {
+	err error
+}
+
+func (c *errorCompressor) Write(_ []byte) (int, error) {
+	return 0, c.err
+}
+
+func (c *errorCompressor) Reset(_ io.Writer) {}
```
smaye81, line 47: Not sure if maybe this should manually panic since there's no way to return an error.

internal/app/server/server.go
```diff
@@ -103,6 +105,7 @@ func createServer(req *v1alpha1.ServerCompatRequest) (*http.Server, error) {
 	mux := http.NewServeMux()
 	mux.Handle(conformancev1alpha1connect.NewConformanceServiceHandler(
 		&conformanceServer{},
+		connect.WithCompression(compression.Zstd, compression.NewZstdDecompressor, compression.NewZstdCompressor),
```
smaye81, line 108: We don't have a way to specify which compression a server should support as part of `ServerCompatRequest`, but maybe we should? Otherwise, we have to specify all of them all the time, which means we can't test the scenario where a client specifies a compression algorithm the server doesn't support. 

Additionally, should we also allow for specifying a client with asymmetric compression (i.e. `WithAcceptCompression`, but not `WithSendCompression`)? Right now, it's all or nothing and the docs actually call this scenario out specifically:

> In particular, Connect encourages asymmetric compression: clients can send uncompressed requests while asking for compressed responses. ...

internal/compression/sentinel.go
jhump: I think it's fine as a no-op. The error will be returned the next time any other method is called, which should be okay.

internal/app/server/server.go
jhump: > We don't have a way to specify which compression a server should support as part of `ServerCompatRequest`

This is a "supported feature". So the server should configure itself to support everything possible. And then all actual supported compression algorithms are enabled via a config file.

jhump: > Otherwise, we have to specify all of them all the time,

I think this is fine and expected.

> which means we can't test the scenario where a client specifies a compression algorithm the server doesn't support.

This will be like having the reference client invoke an undefined entry point or sending intentionally malformed requests: only the reference client will be able to do it, and it will be done with special hand-crafted requests that don't go through the current logic. And this special request would use an intentionally non-existent compression name like "no-such-compression".

Such special requests are not yet implemented in the reference client (and special responses not yet implemented in the reference server). And I think that's totally fine for now.

internal/compression/compression.go
jhump: I don't think these are expensive enough to create to worry about caching. A better place to cache, once the client is processing multiple RPC requests (instead of just one), is to probably cache entire RPC clients, keyed by target URL and options.

jhump (approved):

PR #623, opened 2023-11-01
V2: Add remaining compression algorithms

smaye81:
This adds the remaining compression algorithms to the client and server as specified in the `Compression` enum. So, rounding out the list, conformance tests support:

* brotli
* deflate
* gzip
* snappy
* zstd

### DIFF
--- a/internal/app/client/client.go
+++ b/internal/app/client/client.go
@@ -152,8 +152,8 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
 				compression.NewBrotliDecompressor,
 				compression.NewBrotliCompressor,
 			),
+			connect.WithSendCompression(compression.Brotli),
 		)
-		clientOptions = append(clientOptions, connect.WithSendCompression(compression.Brotli))
 	case v1alpha1.Compression_COMPRESSION_DEFLATE:
 		clientOptions = append(
 			clientOptions,
@@ -162,8 +162,8 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
 				compression.NewDeflateDecompressor,
 				compression.NewDeflateCompressor,
 			),
+			connect.WithSendCompression(compression.Deflate),
 		)
-		clientOptions = append(clientOptions, connect.WithSendCompression(compression.Deflate))
 	case v1alpha1.Compression_COMPRESSION_GZIP:
 		// Connect clients send uncompressed requests and ask for gzipped responses by default
 		// As a result, specifying a compression of gzip for a client indicates it should also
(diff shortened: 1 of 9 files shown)


internal/app/client/client.go
```diff
@@ -143,13 +143,42 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
...
+	case v1alpha1.Compression_COMPRESSION_BR:
+		clientOptions = append(
+			clientOptions,
+			connect.WithAcceptCompression(
+				compression.Brotli,
+				compression.NewBrotliDecompressor,
+				compression.NewBrotliCompressor,
+			),
+		)
+		clientOptions = append(clientOptions, connect.WithSendCompression(compression.Brotli))
```
jhump, line 156: Stylistically, I think this might be better as a single append, instead of two:
```suggestion
		clientOptions = append(
			clientOptions,
			connect.WithAcceptCompression(
				compression.Brotli,
				compression.NewBrotliDecompressor,
				compression.NewBrotliCompressor,
			),
			connect.WithSendCompression(compression.Brotli),
		)
```

internal/app/client/client.go
```diff
@@ -143,13 +143,42 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
...
+	case v1alpha1.Compression_COMPRESSION_DEFLATE:
+		clientOptions = append(
+			clientOptions,
+			connect.WithAcceptCompression(
+				compression.Deflate,
+				compression.NewDeflateDecompressor,
+				compression.NewDeflateCompressor,
+			),
+		)
+		clientOptions = append(clientOptions, connect.WithSendCompression(compression.Deflate))
```
jhump, line 166: Ditto here (and below/throughout)

internal/compression/deflate.go
```diff
@@ -0,0 +1,60 @@
...
+	reader io.ReadCloser
+}
+
+func (c *deflateDecompressor) Read(bytes []byte) (int, error) {
+	return c.reader.Read(bytes)
+}
+func (c *deflateDecompressor) Reset(rdr io.Reader) error {
+	r, ok := c.reader.(flate.Resetter)
+	if !ok {
+		return errors.New("deflate reader is not able to be used as a resetter")
```
jhump, line 36: I assume this should never happen in practice, right? Maybe a comment to that effect so that the reader knows this would truly be exceptional and is not expected to be an actual outcome.

jhump (approved):
Nice!

internal/app/client/client.go
smaye81: Good call. Updated.

internal/compression/deflate.go
smaye81: Yeah it should not. I added a comment there to indicate it's not expected but is just there as a safeguard.

PR #624, opened 2023-11-02
V2: Add process-handling helpers and the client runner

jhump:
This defines some process-related types, for modeling client and server processes.

It also includes a "client runner", which is a wrapper around a client-under-test or reference client process that handles the serialization of requests to the process's stdin and the de-serialization of responses from the process's stdout.

It also adds helpers for reading/writing streams of delimited messages -- where each message is delimited by a four-byte preface that indicates the message's size.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x, 1.21.x]
+        go-version: [1.20.x, 1.21.x]
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
--- a/internal/app/connectconformance/delimited.go
+++ b/internal/app/connectconformance/delimited.go
@@ -1,0 +1,58 @@
+// Copyright 2023 The Connect Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connectconformance
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+
+	"google.golang.org/protobuf/proto"
+)
+
+func readDelimitedMessage[T proto.Message](in io.Reader, msg T) error {
+	var lenBuffer [4]byte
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x, 1.21.x]
```
jhump, line 15: The `go.mod` already indicated this needs Go 1.20. And I'm using some stuff from `os/exec` that was added in Go 1.20. So dropping 1.19.

internal/app/connectconformance/delimited.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+
+	"google.golang.org/protobuf/proto"
+)
+
+func readDelimitedMessage[T proto.Message](in io.Reader, msg T) error {
```
timostamm, line 26: Can't we use the `protodelim` package here? https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/encoding/protodelim

jhump: Yeah, I suppose so. I didn't actually realize that was lurking in there :P
I was instead looking in `protowire` for a streaming way to read a varint, and found nothing. I even [implemented it myself](https://github.com/connectrpc/conformance/pull/607/files#diff-befd9dca9ffdfe6fb049d1f36e9607eab4060e56006c3fcd3285101a823eb18aR339-R367) before deciding it was kinda silly complicated.

I guess if you'd prefer we stick with varint-delimited messages, I can change to that. This is actually using a fixed-length delimiter instead. I guess maybe all runtimes will have some means of easily supporting varint-delimited message streams, so we should be safe to use that?

timostamm: It's fairly new, added in v1.29.

I'm not sure that I'm a fan of varint over a simple fixed size, but the varint-delimitation is available in [C++](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/delimited_message_util.h), [Java](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/AbstractParser#parseDelimitedFrom-java.io.InputStream-), Go, [Swift](https://github.com/apple/swift-protobuf/blob/10295a85cdf2a50ff2125aa5903fe2e18e78e556/Sources/SwiftProtobuf/BinaryDelimited.swift), and [protobuf-es](https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#size-delimited-messages).

Considering that Connect is _very_ likely to be used with protobuf, it seems like a fair choice to use it.

smaye81 (approved):

internal/app/connectconformance/delimited.go
jhump: Okay. I'll come back to this. I have a couple of other PRs stacked on top of this one. So let me get through all those, and then I'll open another PR to change to use varint lengths, and we can all vote on that then.

PR #626, opened 2023-11-03
V2: Update client and server programs to use delimited streaming input

jhump:
This contains several changes to the reference client and reference server.

1. The biggest change is that both of these now read/write delimited messages. This shouldn't be necessary on the server, since it only ever reads and writes one message. However, in practice, I ran into a lot of difficulty consuming all input from the server's stdout: it seems that the `os/exec` package doesn't correctly close the associated stdout writer when it sees EOF on the process's stdout. So trying to use `io.ReadAll` to fully consume the process's stdout just hung forever.

   One potential solution is to use a feature of `exec.Cmd` to have it use an OS pipe for stdout (using `pipe` syscall, instead of creating an in-memory pipe in the Go process). That way, when the process closes its stdout, the runner can see the EOF. However, the docs state that the reader may be closed prematurely when you do this unless you delay calling `Cmd.Wait` until after reaching EOF of the reader. That would require a bit of re-structuring because the code currently calls `Cmd.Wait` eagerly (in a goroutine), just so it can discover when the process terminates and cancel an associated context. So if we used this technique, we'd have an issue in reading the reference server's stderr (for out-of-band error information) because the stderr reader may be closed before we actually can consume all of the data ☹️.

   In any event, it seemed like a reasonable change for consistency, so both client and server can be implemented in the same way, except that the server does not need to loop over the input to EOF but just needs to read one message. Also, there's no risk of the runner just hanging if the server-under-test fails to close its stdout. And it solves the problem because the length prefix means the runner knows exactly how many bytes to read for the server reply, instead of waiting until it sees EOF.
   
2. This also changes the flags management in the reference client and server. They were previously ignoring the args slice passed to them, which means they would always be looking at the flags passed to the test runner when they were spun up in-process. While making this change, I also update these programs so that they won't start if you pass spurious positional command-line args.

3. I've updated the server to include the observed timeout in the payload response. This currently looks at the timeout on the context, but it should probably be changed to instead dig it out of headers. (But that's a bigger pain since the logic is a little different depending on whether the protocol in use is connect or grpc.)

4. Finally, I've updated the server test files to have `.json` extensions and updated the client test files to use consistent ports. The client tests now use port 23456 for HTTP/1 and 23457 for HTTP/2. Also, since the client now accepts multiple requests streamed via stdin, you can do the following to run all of the HTTP/1 test files:

   ```
   go build ./cmd/client && go build ./cmd/server
   ./server -port 23456 -json < testdata/server/connecth1.json &
   cat testdata/client/connecth1* | ./client -json
   kill %1 # stop the server
   ```

   Running the HTTP/2 test files is similar:

   ```
   ./server -port 23457 -json < testdata/server/connecth2.json &
   cat testdata/client/connecth2* | ./client -json
   kill %1 # stop the server
   ```


internal/app/server/impl.go
```diff
@@ -273,13 +275,19 @@ func parseUnaryResponseDefinition(
 }
 
 // Creates request info for a conformance payload.
-func createRequestInfo(headers http.Header, reqs []*anypb.Any) *v1alpha1.ConformancePayload_RequestInfo {
+func createRequestInfo(ctx context.Context, headers http.Header, reqs []*anypb.Any) *v1alpha1.ConformancePayload_RequestInfo {
 	headerInfo := app.ConvertToProtoHeader(headers)
 
+	var timeoutMs *int64
+	if deadline, ok := ctx.Deadline(); ok {
+		timeoutMs = proto.Int64(time.Until(deadline).Milliseconds())
```
smaye81, line 283: Ah nice. I meant to ask about how to handle this value.

smaye81 (approved):
Awesome. Do you think it's worth it to write a few `make` targets to make running these easy? Or if not, we could document them somewhere?

jhump:
> Do you think it's worth it to write a few make targets to make running these easy? Or if not, we could document them somewhere?

That's a good question. I suspect once the test runner is landed, running these binaries this way may not be as useful, so maybe the make targets aren't needed. But I could totally be wrong. If you would like to add targets, seems fine to me.

PR #627, opened 2023-11-03
V2: Some fixes in the test runner machinery

jhump:
In a separate branch, I got the test runner functional. But doing so identified a few other issues in some of the process machinery. So this PR includes fixes to that machinery, so that adding the test runner itself will be a small and very-easy-to-review PR.

### DIFF
--- a/internal/app/connectconformance/client_runner.go
+++ b/internal/app/connectconformance/client_runner.go
@@ -36,29 +36,31 @@ type clientRunner interface {
 	sendRequest(req *conformancev1alpha1.ClientCompatRequest, whenDone func(string, *conformancev1alpha1.ClientCompatResponse, error)) error
 	closeSend()
 	waitForResponses() error
+
+	isRunning() bool
+	stop()
 }
 
 func runClient(ctx context.Context, start processStarter) (clientRunner, error) {
-	ctx, cancel := context.WithCancel(ctx)
 	proc, err := start(ctx, false)
 	if err != nil {
-		cancel() // prevent any context-related leak
 		return nil, err
 	}
-	proc.whenDone(func(_ error) { cancel() })
 	result := &clientProcessRunner{
 		proc:       proc,
-		cancel:     cancel,
 		done:       make(chan struct{}),
 		pendingOps: map[string]func(string, *conformancev1alpha1.ClientCompatResponse, error){},
 	}
+	proc.whenDone(func(_ error) {
+		result.terminated.Store(false)
+	})
 	go result.consumeOutput()
 	return result, nil
 }
 
 type clientProcessRunner struct {
-	proc   *process
-	cancel context.CancelFunc
+	proc       *process
+	terminated atomic.Bool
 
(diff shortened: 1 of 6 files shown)


internal/app/connectconformance/client_runner.go
```diff
@@ -36,29 +36,31 @@ type clientRunner interface {
 	sendRequest(req *conformancev1alpha1.ClientCompatRequest, whenDone func(string, *conformancev1alpha1.ClientCompatResponse, error)) error
 	closeSend()
 	waitForResponses() error
+
+	isRunning() bool
+	stop()
 }
 
 func runClient(ctx context.Context, start processStarter) (clientRunner, error) {
-	ctx, cancel := context.WithCancel(ctx)
```
jhump, line 42: This extra context was unused, so no need dealing with the cancellation of it.

At one point, I plumbed this through to `consumeOutput`, to have it abort when the context was cancelled. But then that would cause a race between it consuming all of the program's output and the process being cancelled (the parent context gets cancelled asynchronously when we observe the program has terminated).

So we rely solely on seeing EOF for `consumeOutput` to return (or for the timeout to elapse), instead of cancelling this context.

internal/app/connectconformance/client_runner.go
```diff
@@ -185,6 +216,7 @@ func (c *clientProcessRunner) consumeOutput() {
 			}
 		case <-time.After(testCaseTimeout):
 			reasonForReturn = errors.New("timed out waiting for result from client")
+			return
```
jhump, line 219: Oof, straight up bug that meant that timeouts weren't working correctly 😝

internal/app/connectconformance/config.go
```diff
@@ -62,7 +62,7 @@ func parseConfig(configFileName string, data []byte) ([]configCase, error) {
 			Path: configFileName,
 		}
 		if err := opts.Unmarshal(data, &config); err != nil {
-			return nil, err
+			return nil, ensureFileName(err, configFileName)
```
jhump, line 65: The protoyaml package tries to report good errors with file and line numbers. However, if it first uses a YAML library to parse the file into an AST. And if that step fails, it returns an error that has no filename in it, which is very confusing for users.

internal/app/connectconformance/process.go
```diff
@@ -70,13 +70,34 @@ func runCommand(command []string) processStarter {
...
 			return nil, err
 		}
-		return (*cmdProcess)(cmd), nil
+		cmdProc := &cmdProcess{
+			cmd:  cmd,
+			done: make(chan struct{}),
+		}
+		go func() {
+			// cmd.Wait can only be called once. So we call it from this goroutine
+			// and then publish the result so it can be read via cmdProc.result()
```
jhump, line 79: This was a fun one. I find it rather strange that `Wait` should only be called once. There's no other way to retrieve the status of the program. So if you have other goroutines that want to watch for when the program terminates, you have to do something like this 🤷

smaye81 (approved):

PR #628, opened 2023-11-03
Add initial implementation for gRPC server

smaye81:
To gain confidence in our conformance tests and to ensure parity with gRPC, we will be running tests against a pure gRPC server (i.e. grpc-go). The server, like the Connect reference server, implements the `ConformanceService` endpoints and behaves in the same way. This PR implements the `Unary` endpoint with the others to follow.

The intent is that the gRPC server (and eventual client) won't stick around forever, so rather than mix the gRPC-specific code (such as dealing with `Metadata`, gRPC errors, and gRPC codecs), the code has been rearranged so that the gRPC logic can live in its own directories and be removed easily.


internal/grpcutil/metadata.go
```diff
@@ -0,0 +1,69 @@
...
+) metadata.MD {
+	md := make(metadata.MD, len(src))
+	for _, hdr := range src {
+		key := strings.ToLower(hdr.Name)
+		md[key] = hdr.Value
+	}
+	return md
+}
+
+// AddHeaderMetadata adds the given proto Headers into the context.
```
jhump, line 53: FWIW, this does not add the given headers to the context. This is actually adding the response headers to the current RPC. Maybe that's a nit, but in my head, the existing wording is confusing since they're not actually stored in the context. (What's actually happening is that the current stream is stored in the context, and `grpc.SetHeader` and `grpc.SetTrailer` pull the stream out of context and store the metadata there, so it can be written to the client later.)

internal/app/grpcserver/server.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"connectrpc.com/conformance/internal/codec"
+	"connectrpc.com/conformance/internal/config"
+	conformancev1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
+	"google.golang.org/grpc"
+)
+
+// Run runs the server according to server config read from the 'in' reader.
+func Run(_ context.Context, _ []string, inReader io.ReadCloser, outWriter io.WriteCloser, _ io.WriteCloser) error {
+	json := flag.Bool("json", false, "whether to use the JSON format for marshaling / unmarshaling messages")
```
jhump, line 33: See what I did in #626 -- instead of ignoring the `args` param (second param, `[]string` type), you need to [use it to parse flags](https://github.com/connectrpc/conformance/pull/626/files#diff-183b1c01db7b47b94d058df3d38c5f457214d8082087a04c107ffb50040e9a64R41-R46). Otherwise, when run in-process, this is actually looking at the flags provided to the test runner instead of the flags passed as argument to this function.

internal/app/grpcserver/server.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+
+	codec := codec.NewCodec(*json)
+
+	// Unmarshal into a ServerCompatRequest
+	req := &conformancev1alpha1.ServerCompatRequest{}
+	if err := codec.Unmarshal(data, req); err != nil {
+		return err
+	}
```
jhump, line 51: Ditto about looking at #626. I changed `Codec` so its very different now, so this can read the message without using `io.ReadAll`.

internal/headers/headers.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package app
```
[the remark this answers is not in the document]
smaye81: Yeah had same thought when I was creating them. Done.

internal/grpcutil/codec.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"errors"
+
+	"google.golang.org/grpc/encoding"
+	"google.golang.org/protobuf/encoding/protojson"
+	proto "google.golang.org/protobuf/proto"
+)
+
+func init() { //nolint:gochecknoinits // gRPC requires this to register a codec.
+	encoding.RegisterCodec(&jsonCodec{})
```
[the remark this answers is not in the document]
smaye81: Meant to ask about the compression actually. OK, cool. I removed the codec stuff for gRPC.

internal/grpcutil/metadata.go
smaye81: Gotcha. Updated the comments.

internal/app/grpcserver/impl.go
```diff
@@ -136,7 +139,8 @@ func asAny(msg proto.Message) (*anypb.Any, error) {
 	if err != nil {
 		return nil, status.Errorf(
 			codes.Internal,
-			fmt.Sprintf("unable to convert message: %s", err.Error()),
+			"unable to convert message: %s",
```
jhump, line 142: aside: It's quite annoying the `status.Errorf` doesn't actually support the "w" directive, for wrapping errors :/

jhump (approved):
LGTM!

internal/app/grpcserver/impl.go
smaye81: Had the same thought. Makes no sense.

PR #630, opened 2023-11-03
V2: Add the test runner: cmd/connectconformance

jhump:
This finally brings it all together. This also adds a single test case to the embedded test suites (which gets expanded to 20 test cases with default config, due to permutations with HTTP versions, protocols, codecs, etc).

### DIFF
--- a/cmd/connectconformance/main.go
+++ b/cmd/connectconformance/main.go
@@ -27,14 +27,12 @@ const (
 	modeFlagName         = "mode"
 	configFlagName       = "conf"
 	knownFailingFlagName = "known-failing"
-	maxServersFlagName   = "max-servers"
 )
 
 type flags struct {
 	mode             string
 	configFile       string
 	knownFailingFile string
-	maxServers       uint
 }
 
 func main() {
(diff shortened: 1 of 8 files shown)


cmd/connectconformance/main.go
```diff
@@ -0,0 +1,115 @@
...
+and test servers with different properties.
+
+A configuration file may be provided which specifies what features the client
+or server under test supports. This is used to filter the set of test cases
+that will be executed. If no config file is indicated, all tests will be run.
+
+A file with a list of known-failing test cases may also be provided. For these
+cases, the test runner reverses its assertions: it considers the test case
+failing to be successful; if the test case succeeds, it is considered a failure.
+(The latter aspect makes sure that the file doesn't is up-to-date and does not
```
smaye81, line 74: nit: i don't think `doesn't` is supposed to be there.

smaye81 (approved):
Pretty sick to see it all working together!

Two minor things I noticed when running it (you may have already seen these, not sure):

1. If you run without putting in a command (like I did at first) it panics. We should probably validate the command was entered.

2. If you run with a bogus command (i.e. `./connectconformance --mode server ./bogus`), it says `no such file or directory` but it seems to run that through all the permutations first and ends up printing that error message a bunch of times.

cmd/connectconformance/main.go
jhump: Fixed

jhump:
> Two minor things I noticed when running it (you may have already seen these, not sure):

Great catches! I obviously didn't play with it enough to exercise these kinds of cases.

Both fixed now.

PR #638, opened 2023-11-06
V2: Implement rest of gRPC server endpoints

smaye81:
This implements the rest of the endpoints for the gRPC server (client streaming, server streaming, and bidi)


internal/app/grpcserver/impl.go
```diff
@@ -67,25 +71,193 @@ func (c *conformanceServiceServer) Unary(
...
+		}
+		reqs = append(reqs, msgAsAny)
+	}
+
+	if err := grpcutil.AddHeaderMetadata(stream.Context(), responseDefinition.ResponseHeaders); err != nil {
+		return err
+	}
+	if err := grpcutil.AddTrailerMetadata(stream.Context(), responseDefinition.ResponseTrailers); err != nil {
+		return err
+	}
```
jhump, line 109: It is only necessary to set headers and trailers through the context for unary RPCs. For all stream RPCs, it is more clear to instead use the `SetHeader` and `SetTrailer` methods on `stream`.

jhump (approved):

internal/app/grpcserver/impl.go
smaye81: Gotcha. Changed all the streaming endpoints do use the `stream` methods. Also, since unary was now the only one using `AddHeader/TrailerMetadata`, I removed it from the `grpcutil` package and just inlined right in the unary endpoint.

PR #644, opened 2023-11-07
V2: Implement rest of gRPC client endpoints

smaye81:
This implements the rest of the endpoints for the gRPC client.


internal/app/grpcclient/impl.go
```diff
@@ -113,24 +115,224 @@ func (i *invoker) unary(
 }
 
 func (i *invoker) serverStream(
-	_ context.Context,
-	_ *v1alpha1.ClientCompatRequest,
+	ctx context.Context,
+	ccr *v1alpha1.ClientCompatRequest,
```
jhump, line 119: image

😹

internal/app/grpcclient/impl.go
```diff
@@ -113,24 +115,224 @@ func (i *invoker) unary(
...
+	if err != nil {
+		return nil, err
+	}
+	// Read headers and trailers from the stream
+	hdr, err := stream.Header()
+	if err != nil {
+		return nil, err
+	}
+	headers := grpcutil.ConvertMetadataToProtoHeader(hdr)
+	trailers := grpcutil.ConvertMetadataToProtoHeader(stream.Trailer())
```
jhump, line 140: You can't call `stream.Trailer` until you've reached the end of the messages. So you need to move this below the `for` `stream.Recv` loop.

internal/app/grpcclient/impl.go
```diff
@@ -113,24 +115,224 @@ func (i *invoker) unary(
...
+		return nil, err
+	}
+
+	// Read headers and trailers from the stream
+	hdr, err := stream.Header()
+	if err != nil {
+		return nil, err
+	}
+	headers := grpcutil.ConvertMetadataToProtoHeader(hdr)
+	trailers := grpcutil.ConvertMetadataToProtoHeader(stream.Trailer())
```
jhump, line 185: Ditto, need to move this below the call to `stream.CloseAndReceive()`.

jhump: Also need to move the call to `stream.Header`. The expected flow for a client-stream is that the headers, response message, and trailers are all sent _after_ the client uploads all requests.

internal/app/grpcclient/impl.go
```diff
@@ -113,24 +115,224 @@ func (i *invoker) unary(
...
+	stream, err := i.client.BidiStream(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// Read headers and trailers from the stream
+	hdr, err := stream.Header()
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
```
jhump, line 242: Might be best off using this defer pattern above for client and server streams, too.

internal/app/grpcclient/impl.go
smaye81: Lol I had the same thought.

smaye81:
@jhump integrated your feedback for calling stream metadata. Also put in a `defer` for the streaming methods too. It did seem to simplify them quite a bit too.

internal/app/grpcclient/impl.go
```diff
@@ -113,24 +115,222 @@ func (i *invoker) unary(
...
+	} else {
+		// If the call was successful, get the returned payloads
+		result.Payloads = append(result.Payloads, resp.Payload)
+	}
+
+	hdr, err := stream.Header()
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
```
smaye81, line 208: Not sure if it's worth it here since it's the last statement in the function 🤷‍♂️

jhump: Probably not. Better to just inline it.

jhump (approved):
Re that defer in `clientStream`: your assessment that it may not be worth using a defer since it's the last thing seems right. Let's inline it.

Otherwise, looks great!

PR #646, opened 2023-11-08
V2: Add verbose flag to test runner

jhump:
This enables verbose output, which can be interesting to see, and also helpful to sanity check that it's loading thing as expected.


smaye81 (approved):
This is great. I actually had a note to ask you about putting this in. Super helpful.

PR #648, opened 2023-11-08
V2: Add Bidi test cases

smaye81:
This adds two test cases for bidi streaming, both testing successful cases. The two cases cover full and half duplex streams.

Note this also adds a few make targets to build the binaries and to run the tests. It also fixes a minor oversight in the bidi endpoint that was not setting response headers or trailers.

To run the tests:

`make runservertests`


Makefile
```diff
@@ -68,6 +68,29 @@ checkgenerate:
...
+	go build -o $(@) cmd/referenceclient/main.go
+
+$(BIN)/referenceserver: Makefile generate
+	go build -o $(@) cmd/referenceserver/main.go
+
+$(BIN)/grpcclient: Makefile generate
+	go build -o $(@) cmd/grpcclient/main.go
+
+$(BIN)/grpcserver: Makefile generate
+	go build -o $(@) cmd/grpcserver/main.go
```
jhump, line 92: It's best to not build a single file. If we want to add code to the `main` package in a different file, this will prevent it from successfully compiling. Instead, just build the whole package.
```suggestion
$(BIN)/connectconformance: Makefile generate
	go build -o $(@) cmd/connectconformance/

$(BIN)/referenceclient: Makefile generate
	go build -o $(@) cmd/referenceclient/

$(BIN)/referenceserver: Makefile generate
	go build -o $(@) cmd/referenceserver/

$(BIN)/grpcclient: Makefile generate
	go build -o $(@) cmd/grpcclient/

$(BIN)/grpcserver: Makefile generate
	go build -o $(@) cmd/grpcserver/
```

Makefile
```diff
@@ -68,6 +68,29 @@ checkgenerate:
 	@# Used in CI to verify that `make generate` doesn't produce a diff.
 	test -z "$$(git status --porcelain | tee /dev/stderr)"
 
+.PHONY: runservertests
+runservertests: $(BIN)/connectconformance $(BIN)/referenceserver
+	$(BIN)/connectconformance --mode server $(BIN)/referenceserver
+
+.PHONY: runclienttests
+runclienttests: $(BIN)/connectconformance $(BIN)/referenceclient
+	$(BIN)/connectconformance --mode client $(BIN)/referenceclient
```
jhump, line 77: We probably want to add these to the existing `test` target, so running tests in CI will also run both of these.

internal/app/referenceserver/impl.go
```diff
@@ -192,6 +192,12 @@ func (s *conformanceServer) BidiStream(
 			responseDefinition = req.ResponseDefinition
 			fullDuplex = req.FullDuplex
 			firstRecv = false
+
+			// If a response definition was provided, add the headers and trailers
+			if responseDefinition != nil {
+				internal.AddHeaders(responseDefinition.ResponseHeaders, stream.ResponseHeader())
+				internal.AddHeaders(responseDefinition.ResponseTrailers, stream.ResponseTrailer())
+			}
```
jhump, line 200: I would move this to the very top. Setting the headers is a no-op if they've already been sent (which happens with the first call to `stream.Send`), so we want to make sure they are set as the first thing.

Makefile
smaye81: I added these to the `all` target so that `make` runs them. I also added them as a separate call in `ci.yaml` so that we could see them as a separate step in the Github Action. It felt strange to mix them with the `make test` target. Even though they're tests, it felt strange to mix them with the `make test` target since they're a totally different kind of tests. Not a big deal to me though. If you'd rather I put them in `make test` I can do that too, just lmk.

internal/app/referenceserver/impl.go
smaye81: This is actually only run once when we receive the first message (and before any sends). I _thiink_ it should be OK where it is based on what you mentioned, but let me know if I'm missing something.

jhump: I wasn't saying it's in the wrong place. I was suggesting moving to the top because that's a more intuitive place for it IMO. Right now, you have to look closely at the code to decide if this is a bug or not. Whereas if it were at the top, it would be obviously correct.

Actually, I just spotted a problem, so it does actually _need_ to move: if the request stream has _zero_ messages, this will never get called.

jhump: Oof, never mind, I'm an idiot. 🤦 

You have to read the first request in order to know what headers/trailers to set. My bad.

jhump: Homer hiding

jhump (approved):

PR #649, opened 2023-11-09
V2: Proto updates to handle test cases for client certs and max message sizes

jhump:
Hopefully these changes are reasonably clear. They are unused, so I will next start opening PRs to make use of this stuff in the test runner.

I can also make the necessary changes to the reference client and server in subsequent PRs. (I still need to add TLS stuff to the reference client, so I will add the stuff for client certs and max message size then.)


proto/connectrpc/conformance/v1alpha1/server_compat.proto
```diff
@@ -38,6 +38,17 @@ message ServerCompatRequest {
...
+  // If non-empty, the clients will use certificates to authenticate
+  // themselves. This value is a PEM-encoded cert that should be
+  // trusted by the server. When non-empty, the server should require
+  // that clients provide certificates and they should validate that
+  // the certificate presented is valid.
+  //
+  // This will always be empty if use_tls is false.
+  bytes client_tls_cert = 5;
+  // If non-zero, indicates the maximum size in bytes for a message.
+  // If the server sends anything larger, the client should reject it.
```
smaye81, line 50: Think this should be `If the client sends anything larger, the server should reject it` right?

smaye81 (approved):

proto/connectrpc/conformance/v1alpha1/client_compat.proto
```diff
@@ -63,6 +82,12 @@ message ClientCompatResponse {
     ClientResponseResult response = 2;
     ClientErrorResult error = 3;
   }
+
+  // This field is used only be the reference client, and it can be used
```
jhump, line 86:
```suggestion
  // This field is used only by the reference client, and it can be used
```

proto/connectrpc/conformance/v1alpha1/server_compat.proto
jhump: Yes sir, indeed.
```suggestion
  // If the client sends anything larger, the server should reject it.
```

PR #650, opened 2023-11-10
Test cases for Bidi errors

smaye81:
This adds two test cases for errors returned from bidi endpoints for both full and half duplex streams.


jhump (approved):
Nice!

PR #653, opened 2023-11-10
V2: Support TLS and max message size in client; add TLS test case

jhump:
In order to get the TLS test case to work, I had to make some fixes in the server side.

### DIFF
--- a/internal/app/referenceclient/client.go
+++ b/internal/app/referenceclient/client.go
@@ -30,6 +30,7 @@ import (
 	"connectrpc.com/conformance/internal/gen/proto/connect/connectrpc/conformance/v1alpha1/conformancev1alpha1connect"
 	v1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
 	"connectrpc.com/connect"
+	"github.com/quic-go/quic-go/http3"
 	"golang.org/x/net/http2"
 )
 
@@ -92,13 +93,17 @@ func Run(ctx context.Context, args []string, inReader io.ReadCloser, outWriter, 
 // Connect error returned from calling an RPC. Any error (i.e. a Connect error) that _is_ returned from
 // the actual RPC invocation will be present in the returned ClientResponseResult.
 func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.ClientResponseResult, error) {
+	tlsConf, err := createTLSConfig(req)
+	if err != nil {
+		return nil, err
+	}
 	var scheme string
-	if req.ServerTlsCert != nil {
+	if tlsConf != nil {
 		scheme = "https://"
 	} else {
 		scheme = "http://"
 	}
-	urlString := scheme + net.JoinHostPort(req.Host, strconv.FormatUint(uint64(req.Port), 10))
+	urlString := scheme + net.JoinHostPort(req.Host, strconv.Itoa(int(req.Port)))
 	serverURL, err := url.ParseRequestURI(urlString)
 	if err != nil {
 		return nil, errors.New("invalid url: %s" + urlString)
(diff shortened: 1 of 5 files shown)


internal/app/referenceserver/impl.go
```diff
@@ -55,8 +55,8 @@ func (s *conformanceServer) Unary(
 		[]*anypb.Any{msgAsAny},
 	)
 	if connectErr != nil {
-		internal.AddHeaders(req.Msg.ResponseDefinition.ResponseHeaders, connectErr.Meta())
-		internal.AddHeaders(req.Msg.ResponseDefinition.ResponseTrailers, connectErr.Meta())
+		internal.AddHeaders(req.Msg.ResponseDefinition.GetResponseHeaders(), connectErr.Meta())
+		internal.AddHeaders(req.Msg.ResponseDefinition.GetResponseTrailers(), connectErr.Meta())
```
jhump, line 59: This was "just in case", since I had played around with `buf curl` and an empty request to troubleshoot TLS issues. If the error from `parseUnaryResponseDefinition` is that it is complaining that no response definition is provided, these would cause the server to panic and fail to actually report that error.

internal/app/referenceserver/server.go
```diff
@@ -111,6 +111,9 @@ type stdHTTPServer struct {
 }
 
 func (s *stdHTTPServer) Serve() error {
+	if s.svr.TLSConfig != nil {
+		return s.svr.ServeTLS(s.lis, "", "")
+	}
```
jhump, line 116: Doh! This is necessary for the HTTP server to actually be a secure server. The empty strings are for the cert and key, but they can be blank if they are instead provided in the server's `TLSConfig` field (which they are in this case).

internal/app/referenceserver/server.go
```diff
@@ -191,9 +194,6 @@ func createServer(req *v1alpha1.ServerCompatRequest, listenAddr string) (httpSer
 
 // newH1Server creates a new HTTP/1.1 server.
 func newH1Server(handler http.Handler, listenAddr string, tlsConf *tls.Config) (httpServer, error) {
-	if tlsConf != nil {
-		tlsConf.NextProtos = []string{"http/1.1"}
-	}
```
jhump, line 196: This was basically a no-op, so I removed it. It's the _client_ that proposes the protocols, so this was important to set on the client to make sure that HTTP/1.1 or HTTP/2 gets used. It's not necessary on the server.

internal/tls.go
```diff
@@ -60,7 +60,7 @@ func NewClientTLSConfig(caCert, clientCert, clientKey []byte) (*tls.Config, erro
 
 	return &tls.Config{
 		RootCAs:      caCertPool,
-		Certificates: []tls.Certificate{certPair},
+		Certificates: certs,
```
jhump, line 63: Oops, this should be nil/empty slice if there is no client cert.

internal/tls.go
```diff
@@ -138,7 +138,7 @@ func NewServerCert() (tls.Certificate, []byte, error) {
 		IPAddresses: []net.IP{net.IPv6loopback, net.IPv4(127, 0, 0, 1)},
 		DNSNames:    []string{"localhost"},
 	}
-	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, priv.PublicKey, priv)
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
```
jhump, line 141: Another subtle mistake: we need to pass a the _address_ of `priv.PublicKey`, because it accepts a value of type `*rsa.PublicKey` but was instead getting a value of type `rsa.PublicKey`.

internal/app/referenceclient/client.go
```diff
@@ -109,16 +114,39 @@ func invoke(ctx context.Context, req *v1alpha1.ClientCompatRequest) (*v1alpha1.C
 	var transport http.RoundTripper
 	switch req.HttpVersion {
 	case v1alpha1.HTTPVersion_HTTP_VERSION_1:
-		transport = &http.Transport{}
+		if tlsConf != nil {
+			tlsConf.NextProtos = []string{"http/1.1"}
```
jhump, line 118: This forces the client to use HTTP 1.1 with TLS. Without this, since `net/http` supports both HTTP 1.1 _and_ HTTP/2, it would end up trying to negotiate HTTP/2. But the `NextProtos` slice is what defines the set of supported protocols. (Same goes for why we set this to `[]string{"h2"}` below, to make sure we don't unintentionally negotiate HTTP 1.1 with the server.)

smaye81: `NextProtos` is such an unfortunate name. I kept thinking 'why does TLS care about proto files'

smaye81 (approved):

PR #654, opened 2023-11-10
V2: Create config that represents capabilities of the reference implementations

jhump:
This uses the configs to enable a greater number of test cases in `make runconformance` (up to 288 cases/permutations now!). It also runs tests against the gRPC impls as part of `make runconformance`.

This also meant fixing some small issues I found along the way:

1. In order to enable gzip support with gRPC, you have to import their encoding/gzip package.
2. The zstd compression implementation complained if a decoder was used for more than one request. Strangely, unlike all of the other implementations, a zstd decoder cannot be re-used, not even via `Reset`, after it has been `Close`d. 🤷 
3. The logic in the gRPC client to translate a `status.Error` to the error proto in the `ClientCompatResponse` wasn't working. I didn't dig much into exactly why, but I did realize that there was a much simpler way to translate since the error proto and the gRPC status proto are nearly identical. When I simplified the code, the bug went away. (The bug was that error details were getting dropped for some reason.)

This builds on top of #653.

### DIFF
--- a/internal/grpcutil/errors.go
+++ b/internal/grpcutil/errors.go
@@ -16,11 +16,8 @@ package grpcutil
 
 import (
 	v1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
-	"connectrpc.com/connect"
 	statuspb "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc/status"
-	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/types/known/anypb"
 )
 
 // ConvertProtoToGrpcError converts a proto Error into a gRPC error.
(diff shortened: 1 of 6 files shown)


internal/grpcutil/errors.go
```diff
@@ -42,30 +39,11 @@ func ConvertGrpcToProtoError(err error) *v1alpha1.Error {
...
 		return nil
 	}
-	status, ok := status.FromError(err)
-	if !ok {
-		// If the given error is not a gRPC error, return unknown
-		return &v1alpha1.Error{
-			Code:    int32(connect.CodeUnknown),
-			Message: "provided error is not a gRPC error",
-		}
+	stat, _ := status.FromError(err)
```
smaye81, line 42: Do you think it's worth it to check the `ok` return value here?  Just in case we couldn't convert this to a status? Maybe that will never happen in practice I guess.

smaye81 (approved):

internal/grpcutil/errors.go
jhump: When `ok` is false, it still returns a status -- it just returns one with an unknown code and a message set to `err.Error()`.

PR #655, opened 2023-11-10
Add basic success and error tests

smaye81:
This adds basic success and error tests for the remaining rpc types (unary, server stream, and client stream (error))

### DIFF
--- a/internal/app/connectconformance/testsuites/basic.yaml
+++ b/internal/app/connectconformance/testsuites/basic.yaml
@@ -14,7 +14,7 @@ testCases:
       responseDefinition:
         responseHeaders:
         - name: x-custom-header
-          value: ["foo", "bar","baz"]
+          value: ["foo","bar","baz"]
         responseData: "dGVzdCByZXNwb25zZQ=="
         responseTrailers:
         - name: x-custom-trailer
@@ -55,7 +55,7 @@ testCases:
       responseDefinition:
         responseHeaders:
         - name: x-custom-header
-          value: ["foo", "bar","baz"]
+          value: ["foo","bar","baz"]
         error:
           code: 13
           message: "unary failed"
@@ -95,7 +95,7 @@ testCases:
       responseDefinition:
         responseHeaders:
         - name: x-custom-header
-          value: ["foo", "bar","baz"]
+          value: ["foo","bar","baz"]
         responseData: "dGVzdCByZXNwb25zZQ=="
         responseTrailers:
         - name: x-custom-trailer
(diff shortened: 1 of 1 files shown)


internal/app/connectconformance/testsuites/connect_only.yaml
```diff
@@ -0,0 +1,101 @@
...
+    streamType: STREAM_TYPE_UNARY
+    requestHeaders:
+    - name: X-Conformance-Test
+      value: ["Value1","Value2"]
+    requestMessages:
+    - "@type": type.googleapis.com/connectrpc.conformance.v1alpha1.UnimplementedRequest
+  expectedResponse:
+    error:
+      code: 12
+      message: "connectrpc.conformance.v1alpha1.ConformanceService.Unimplemented is not implemented"
```
smaye81, line 101: @jhump We probably don't want to validate the actual error message text here since an implementation can put whatever they want right? I think we'd only want to validate the code, but I don't think we have a way to do that.

internal/app/referenceclient/impl.go
```diff
@@ -333,6 +339,36 @@ func (i *invoker) bidiStream(
...
+		return nil, err
+	}
+
+	request := connect.NewRequest(ur)
+
+	var protoErr *v1alpha1.Error
+
+	// Invoke the Unary call
+	_, err := i.client.Unimplemented(ctx, request)
+	if err != nil {
```
jhump, line 358: This seems a little odd that this doesn't do the same thing as `Unary`. In fact, you could make that method generic (parameterized by the request/response types), so the exact same function could be used for _all_ unary RPCs (admittedly, there are only two of them).

jhump (approved):

internal/app/referenceclient/impl.go
smaye81: I don't think you can make a struct method genericized, right? I tried to implement what you mentioned and kept running into issues (unless I'm missing an easiser way).

jhump: Indeed, it can't be a method. It has to be a function. So you'd have to pass `i` as a normal parameter instead of as a method receiver.

In any event, this was a non-blocking comment. I guess a simpler option is to just return an error instead of an RPC response if `err == nil`.

smaye81: I think we can get even simpler than that. The `ConvertErrorToProtoError` function just returns `nil` if `err == nil`, so we can just set whatever that returns on the `Error` property and return. The test will fail if the error is nil in that case then.

PR #662, opened 2023-11-13
V2: Add support for TLS client certs and message limits to the test runner

jhump:
This also adds a single, simple test case for TLS client cert support (basically a copy of the existing TLS test case).

I had to add a couple of fields to the config proto -- apparently I missed that in previous PR that touched the proto schema.


internal/app/connectconformance/test_case_library.go
```diff
@@ -85,6 +85,21 @@ func newTestCaseLibrary(
...
 
 func (lib *testCaseLibrary) expandSuite(suite *conformancev1alpha1.TestSuite, configCases map[configCase]struct{}) error {
+	if suite.ReliesOnTlsClientCerts && !suite.ReliesOnTls {
+		return fmt.Errorf("suite %q is misconfigured: it relies on TLS client certs but not TLS", suite.Name)
+	}
+	if suite.ReliesOnConnectGet && !only(suite.RelevantProtocols, conformancev1alpha1.Protocol_PROTOCOL_CONNECT) {
+		return fmt.Errorf("suite %q is misconfigured: it relies on Connect GET support, but has unexpected relevant protocols: %v", suite.Name, suite.RelevantProtocols)
+	}
+	if suite.ConnectVersionMode == conformancev1alpha1.TestSuite_CONNECT_VERSION_MODE_IGNORE && !only(suite.RelevantProtocols, conformancev1alpha1.Protocol_PROTOCOL_CONNECT) {
+		return fmt.Errorf("suite %q is misconfigured: it relies on Connect GET support, but has unexpected relevant protocols: %v", suite.Name, suite.RelevantProtocols)
```
smaye81, line 95: Is this the right error message? I think we'd want something here about ignoring Connect version mode, right?

internal/app/connectconformance/test_case_library.go
```diff
@@ -85,6 +85,21 @@ func newTestCaseLibrary(
...
+	}
+	if suite.ConnectVersionMode == conformancev1alpha1.TestSuite_CONNECT_VERSION_MODE_REQUIRE && !only(suite.RelevantProtocols, conformancev1alpha1.Protocol_PROTOCOL_CONNECT) {
+		return fmt.Errorf("suite %q is misconfigured: it ignores Connect Version headers, but has unexpected relevant protocols: %v", suite.Name, suite.RelevantProtocols)
```
smaye81, line 98: Should this be `...it requires Connect Version headers...`?

smaye81 (approved):

internal/app/connectconformance/test_case_library.go
jhump: Yep, thank you for catching that. Copy+pasta error on my part.

PR #663, opened 2023-11-14
Automate adding issues and PRs to project board

chrispine:
This introduces automation for adding issues and PRs to an org-wide project board. The goal is to have a clear view into incoming issues/PRs so that we know none can fall through the cracks.

This was first tested in examples-go: https://github.com/connectrpc/examples-go/pull/145

### DIFF
--- a/.github/add-to-project.yaml
+++ b/.github/add-to-project.yaml
@@ -1,0 +1,28 @@
+name: Add issues and PRs to project
+
+on:
+  issues:
+    types:
+      - opened
+      - reopened
+      - transferred
+  pull_request:
+    types:
+      - opened
+      - reopened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


nicksnyder (approved):

jhump:
@chrispine, for all of these recent governance/management PRs, do you mind also opening PRs that make these same changes in the `v2` branch?

chrispine:
@jhump Two things happened at about the same time:

- I added this project automation, which is mainly to make sure external contributions/issues don't fall through the cracks (by adding them to the board I shared earlier today).
- Akshay made some governance/consistency changes to all of the repos in this org.

Are you asking for both of those, or only the second one?

nicksnyder:
I think/assume the `v2` branch is going to become the default branch, so we need to get the add-to-project.yaml workflow added to that branch (because everything currently on main is going to get nuked).

chrispine:
Here it is: https://github.com/connectrpc/conformance/pull/675

PR #667, opened 2023-11-15
V2: Autopopulate expected responses

smaye81:
This autopopulates the expected response information based on what is provided for requests in the test yaml files. This allows for eliminating a lot of boilerplate in the yaml files for straight-forward tests for all stream types. 

The logic looks at the request information and then dynamically builds a `ClientResponseResult` based on what is being sent to the server.

This also allows for overriding the autopopulation by explicitly specifying an `expectedResult` in the yaml file. This can be helpful for one-off or out-of-the-ordinary requests, such as the Unimplemented call or evaluating gRPC response trailers.

Note that the above could also be built into the autopopulation, but this base path might be a better place to start to gauge whether this is a good idea.


smaye81:
@jhump I realized that we could autopopulate the entire `expectedResult` just by evaluating the request information. I also added in the ability to override the autopopulation via the yaml file. Let me know what you think. If you think it's too much or won't apply to other, more complex tests, I can remove / scale it back.

internal/app/connectconformance/server_runner.go
```diff
@@ -26,8 +26,145 @@ import (
...
+// by examining the requests we sent.
+func populateExpectedRequests(testCase *conformancev1alpha1.TestCase) error {
+	// If an expected response was already provided, return and use that.
+	// This allows for overriding this function with explicit values in the yaml file.
+	if testCase.ExpectedResponse != nil {
+		return nil
+	}
+	if len(testCase.Request.RequestMessages) == 0 {
+		return errors.New("at least one request is required")
+	}
```
jhump, line 42: I don't think we want this constraint. We want to be able to test client and bidi streams where there are no upload messages.

Maybe add a TODO and then we can follow-up with updates to the server impls to produce (and the code below to expect) a single response message in this situation, where the response data value is some fixed string (such as `"no response definition"`) and whose request info will still be present, but we expect it to indicate zero request messages.

In the future, we could come up with a very simple scheme for how the server can read a minimal response definition from request headers.

(Just riffing now...)

Instead of a full response definition, it could be something very simple like:
```
x-response-definition: foo:bar, fu:baz, 3:abcdef, 9:some error message
```
Where the value is a simple comma-delimited string with four components:
1. Pipe-delimited custom headers. Each header is in `name:value` form. So `foo:bar|foo:baz|frob:nitz` is equivalent to
   ```yaml
   [
      {name: "foo", value: ["bar", "baz"]},
      {name: "frob", value:["nitz"]}
   ]
   ```
   To keep parsing simple, the header values may not contain pipes or commas.
2. Pipe-delimited custom trailers.
3. Response messages in the form `count:data`. The count must be zero or one for client-stream endpoints, depending on if next field indicates an error or not. So `3:foo` will send back three response messages where the `data` field is `"foo"` (as bytes). The data value may not contain commas (to keep parsing of the definition simple).
4. Error in the form `code:message`. This string will be the simple value `"ok"` to mean no error. The code portion will be the numeric value of the gRPC code to use.

internal/app/connectconformance/server_runner.go
```diff
@@ -26,8 +26,145 @@ import (
 	"connectrpc.com/conformance/internal"
 	conformancev1alpha1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1alpha1"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
+// populateExpectedRequests populates the requests we expected to get back from the server
+// by examining the requests we sent.
+func populateExpectedRequests(testCase *conformancev1alpha1.TestCase) error {
```
jhump, line 34: I think there's enough logic here that we want a test. Otherwise, there's the risk of a bug that means we may fail to catch certain issues with the response.

internal/app/connectconformance/server_runner.go
```diff
@@ -26,8 +26,145 @@ import (
...
+				// For server streams, only one request was sent. All request information specified
+				// should only be echoed back in the first response
+				if idx == 0 {
+					expected.Payloads[idx].RequestInfo = &conformancev1alpha1.ConformancePayload_RequestInfo{
+						RequestHeaders: testCase.Request.RequestHeaders,
+						Requests:       testCase.Request.RequestMessages,
+					}
+				}
+			case conformancev1alpha1.StreamType_STREAM_TYPE_HALF_DUPLEX_BIDI_STREAM:
+				// For half-duplex streams, all request information should be echoed back in all
```
jhump, line 95: > all request information should be echoed back in all

Hmmm. I didn't realize it was implemented this way. There's no reason to assert this information repeatedly. It should only be in the first response. There's even a bit of logic in the test runner, that checks for this, that only care about the first message:
https://github.com/connectrpc/conformance/blob/v2/internal/app/connectconformance/results.go#L338

(That's for things like headers and query string params, since they can never change. The request messages can change from one to the next, which is expected to be the case for full-duplex bidi streams.)

jhump: Changing this means you can collapse this into the above block, since the logic would then be the same as for server streams.

internal/app/connectconformance/server_runner.go
jhump: Also, this should probably be called `populateExpectedResponse`, since the name of the field it populates is `ExpectedResponse`, not `ExpectedRequests`.

internal/app/connectconformance/server_runner.go
```diff
@@ -26,8 +26,145 @@ import (
...
+
+	switch testCase.Request.StreamType {
+	case conformancev1alpha1.StreamType_STREAM_TYPE_FULL_DUPLEX_BIDI_STREAM,
+		conformancev1alpha1.StreamType_STREAM_TYPE_HALF_DUPLEX_BIDI_STREAM,
+		conformancev1alpha1.StreamType_STREAM_TYPE_SERVER_STREAM:
+
+		// First, find the response definition that the client instructed the
+		// server to return
+		var def *conformancev1alpha1.StreamResponseDefinition
+		if testCase.Request.StreamType == conformancev1alpha1.StreamType_STREAM_TYPE_SERVER_STREAM {
```
jhump, line 54: Instead of a conditional here, you could use `req.UnmarshalNew` (instead of `UnmarshalTo`) and then convert the result to an interface you can use to get the value. Something like so:
```go
concreteReq, err := req.UnmarshalNew()
if err != nil {
    return err
}
type streamResponseDefiner interface {
    GetResponseDefinition() *conformancev1alpha1.StreamResponseDefinition
}
//nolint:forcetypeassert
def = ur.(streamResponseDefiner).GetResponseDefinition()
```

And you can do similar for the types where `GetResponseDefinition()` returns a `UnaryResponseDefinition` below.

internal/app/connectconformance/server_runner.go
```diff
@@ -26,8 +26,145 @@ import (
...
+				// For a full duplex stream, the first request information should be echoed back in the first
+				// payload. The second should be echoed back in the second payload, etc. (i.e. a ping pong interaction)
+				expected.Payloads[idx].RequestInfo = &conformancev1alpha1.ConformancePayload_RequestInfo{
+					RequestHeaders: testCase.Request.RequestHeaders,
+					Requests:       []*anypb.Any{testCase.Request.RequestMessages[idx]},
+				}
+			}
+		}
+		testCase.ExpectedResponse = expected
+	case conformancev1alpha1.StreamType_STREAM_TYPE_UNARY, conformancev1alpha1.StreamType_STREAM_TYPE_CLIENT_STREAM:
```
jhump, line 112: These cases are big enough and reasonably self-contained, so I think it would be better to factor them out into two methods -- like a `populateExpectedStreamResponse` and `populateExpectedUnaryResponse`.

jhump (commented):
Very cool.

internal/app/connectconformance/server_runner.go
smaye81: I left a few TODOs in the code here to address this. Both places may be redundant, but at least we'll know where to make sure to cover our bases.

internal/app/connectconformance/server_runner.go
smaye81: Renamed and added tests. I apologize in advance for the size of it. Not sure if there's a better way. There are a lot of cases to cover and a lot of structs to create to do so 😬

internal/app/connectconformance/server_runner.go
smaye81: Fixed and I also updated the server impls to handle this better. Now, the request information (more accurately -- anything that doesn't change from request to request) is only echoed back in the first response.

internal/app/connectconformance/server_runner.go
smaye81: Awesome. Updated. I added a check on the type assert but can remove if you think it's unnecessary.

internal/app/connectconformance/server_runner.go
smaye81: Moved out into their own functions, but I kept the entry point for `populateExpectedResponse` so it didn't clutter up the `runTestCasesForServer` function. I can remove that also too, so just lmk.

internal/app/connectconformance/server_runner.go
```diff
@@ -30,140 +30,179 @@ import (
...
+	}
+
+	// TODO - Need to define this better in the protos and tests as to how services should
+	// behave if no responses are specified. The behavior right now differs for unary vs. streaming
+	// If no responses are specified for unary, the service will still return a response with the
+	// request information inside (but none of the response information since it wasn't provided)
+	// But streaming endpoints don't return a single response and instead return responses via sending
+	// on a stream. But if no responses are specified in the request, the streams don't send anything outbound
+	// so there's no way to relay this to a client. So right now, streaming endpoints simply expect an empty
+	// ClientResponseResult if no response definition is provided
```
jhump, line 57: Interesting point. TBH, we can probably get away with doing nothing -- if the response definition says to send back no messages, then the request doesn't get echoed back. And that's probably okay since we'll have coverage of the request handling and it being echoed back from so many other test cases where there is 1+ response messages.

jhump (approved):
Nice!

PR #669, opened 2023-11-16
V2: Update proto version to v2

jhump:
We were using "v1alpha1", however that can't be right since the `main` branch already uses a package named "v1": https://github.com/connectrpc/conformance/tree/main/proto/connectrpc/conformance/v1.
(My bad, I didn't realize how the existing protos were structured/named, or I wouldn't have suggested "v1alpha1" initially...)

So this changes it to "v2". The real question: do we hedge our bets and instead call it "v2alpha1"? Or do we just hold off to actually publish (by merging this to main and pushing to the BSR) until we're comfortable enough that "v2" is the right name?

I also went ahead and did `git rm _legacy`, just because a `git grep` found some references to "v1alpha1" in there, too. I didn't quite understand why they were there, so I figured it was better to just wipe that whole folder since AFAICT we haven't needed it (and can always get all of that code from the `main` branch if ever we do need any of it).

When we do merge `v2` branch to `main`, I think we should also create a `v1` branch at that time that points to the last commit before the merge. That way we have an easy way to reference the old code.


smaye81:
I was originally thinking the `v1` on main didn't matter since we didn't publish those to the BSR yet, but I guess those protos have been available for anyone to use from the repo, so we can't make breaking changes to them now.

My vote would be to hold off until we actually publish and then just use `v2`, but no biggie either way.

smaye81 (approved):

timostamm:
@smaye81 FYI, the v1 protos are on the BSR: https://buf.build/connectrpc/conformance/

jhump:
> FYI, the v1 protos are on the BSR

Oof, that was done due to a miscommunication. I thought we removed them. https://bufprivate.slack.com/archives/C03FFSJELH4/p1697489587709479?thread_ts=1697480343.693549&cid=C03FFSJELH4

smaye81:
> > FYI, the v1 protos are on the BSR
> 
> Oof, that was done due to a miscommunication. I thought we removed them. https://bufprivate.slack.com/archives/C03FFSJELH4/p1697489587709479?thread_ts=1697480343.693549&cid=C03FFSJELH4

Ah yeah I remember when they got published, but I thought they were removed also.

PR #672, opened 2023-11-17
V2: Add ability to define raw request and response in test cases

jhump:
This adds the things we discussed the other day: a way to define a raw HTTP request or response, for testing edge cases and handling of unusual/arbitrary requests and responses.

It also addresses a couple of other TODOs that we've discussed in the past: it provides a way to direct the client to cancel the RPC; and it adds a field where the client can record a number of requests that could not be sent due to the server finishing the RPC before half-closing.


proto/connectrpc/conformance/v2/client_compat.proto
```diff
@@ -73,6 +74,38 @@ message ClientCompatRequest {
   // For client- or bidi-streaming requests, this delay should be
   // applied before each request sent.
   uint32 request_delay_ms = 18;
+  // If present, the client should cancel the RPC instead of
+  // allowing to complete normally.
+  Cancel cancel = 19;
+  message Cancel {
+    // These fields determine the timing of cancellation.
+    // If none are present, the client should cancel immediately
+    // after all request messages are sent the stream half-closed
```
smaye81, line 83: Can you elaborate on what you mean by `half-close`? Does that mean closing the send side for client/bidi streams and closing the receive side for server streams?

Also, just to confirm I'm following:

- `before_half_close` example would be a client or bidi stream sends all its requests and then cancels (in Go parlance, we'd cancel the context) and then closes the send side? 
- `after_half_close_ms` example would be a server stream sends the request, closes the send side of the stream, waits for `x` ms and then cancels?

proto/connectrpc/conformance/v2/client_compat.proto
```diff
@@ -73,6 +74,38 @@ message ClientCompatRequest {
...
+      // sent. This applies only to client and bidi stream RPCs.
+      google.protobuf.Empty before_half_close = 1;
+      // When present, the client should delay for this many
+      // milliseconds after half-closing the stream and then
+      // cancel.
+      uint32 after_half_close_ms = 2;
+      // When present, the client should cancel right after
+      // reading this number of response messages from the stream.
+      // This applies only to server and bidi stream RPCs.
+      int32 after_num_responses = 3;
```
smaye81, line 97: Should this be `uint32`?

proto/connectrpc/conformance/v2/client_compat.proto
```diff
@@ -73,6 +74,38 @@ message ClientCompatRequest {
...
   uint32 request_delay_ms = 18;
+  // If present, the client should cancel the RPC instead of
+  // allowing to complete normally.
+  Cancel cancel = 19;
+  message Cancel {
+    // These fields determine the timing of cancellation.
+    // If none are present, the client should cancel immediately
+    // after all request messages are sent the stream half-closed
+    // (as if the after_half_close_ms field were present and zero).
+    oneof cancel_timing {
```
smaye81, line 85: Should we also have the ability to specify never half-closing and _only_ cancelling? Not sure if the expected behavior would be any different.

proto/connectrpc/conformance/v2/client_compat.proto
smaye81, line 97: It makes sense that for server and half-duplex streams, the send side would be closed here. But, for full-duplex bidi streams, does this imply cancelling and half-closing? Because if we say for example, 'cancel after reading 1 response on a full-duplex stream', the send side may not have been closed? Hopefully this makes sense.

proto/connectrpc/conformance/v2/client_compat.proto
```diff
@@ -73,6 +74,38 @@ message ClientCompatRequest {
...
+  }
+
+  // The following field is only used by the reference client. If
+  // you are implementing a client under test, you may ignore it
+  // or respond with an error if the client receives a request where
+  // it is set.
+  //
+  // The raw request may be specified instead of the above group
+  // of fields for defining the request to send.
+  RawHTTPRequest raw_request = 20;
```
smaye81, line 108: Should we group this with the first 8 fields above since it also represents a field that shouldn't be explicitly set by tests? Just a thought 🤷‍♂️

proto/connectrpc/conformance/v2/service.proto
```diff
@@ -243,3 +262,75 @@ message Header {
   string name = 1;
   repeated string value = 2;
 }
+
+// RawHTTPRequest models a raw HTTP request. This can be used to craft
+// custom requests with odd properties (including certain kinds of
+// malformed requests) to test edge cases in servers.
+message RawHTTPRequest {
+  string verb = 1;
```
smaye81, line 270: Could we model this with an enum?

proto/connectrpc/conformance/v2/client_compat.proto
jhump: "Half-close" means the client closes the request-ending side. The other "half", the server sending responses, remains open until the stream is complete.

In grpc-go, this operation is called "close send". In connect-go, it's called "close request", In the HTTP/2 specs, this state is called "half-closed".

> `before_half_close` example would be a client or bidi stream sends all its requests and then cancels (in Go parlance, we'd cancel the context) and then closes the send side?

Mostly right. After cancelling, there should be no need to close the send side: cancelling should implicitly close everything.

> `after_half_close_ms` example would be a server stream sends the request, closes the send side of the stream, waits for `x` ms and then cancels?

Exactly.

jhump: I guess I can rename these `before_close_send` and `after_close_send_ms`?

proto/connectrpc/conformance/v2/client_compat.proto
jhump: Sure.

proto/connectrpc/conformance/v2/client_compat.proto
jhump: My expectation is that really depends on the client library. At a low-level, you should _never_ have to half-close after cancelling: cancelling implicitly closes the whole thing. So that's the behavior I imagined for `before_half_close`. In HTTP/2 terms, once the stream is cancelled, nothing else is ever sent on it. When you half-close a stream, a final frame with "end of stream" bit set is sent to tell the server the stream is half-closed. So half-closing on the client-side shouldn't actually do anything on the wire if cancel actually works.

In practice however, it is possible that a particular implementation may want client code to use half-close/close-send, even after a cancellation, to ensure there are no resource leaks (like if there are other resources associated with the stream that need to be explicitly freed vs. auto-freed on cancellation).

Hopefully for our current implementations, half-close isn't necessary after cancel, and that's how I'd recommend we implement them.

proto/connectrpc/conformance/v2/client_compat.proto
jhump: It makes sense. For bidi, I'd expect this to take priority over the number of requests. So if we cancel after 3 responses, but we've only sent 3 out 5 requests, we'd cancel and then record that 2 requests could not be sent (in other new field).

As far as "imply cancelling and half-closing": as mentioned above, cancelling implicitly closes the whole stream. So... kind of, yes. Except that the code should not actually need to call the half-close/close-send method: it just cancels.

proto/connectrpc/conformance/v2/client_compat.proto
jhump: No, it _should_ be explicitly set by tests. The way you write a test case that uses a raw request is to explicitly set it in the test case _instead of_ the previous group of fields.

proto/connectrpc/conformance/v2/service.proto
jhump: We could. I don't know that it's useful, and that would prevent a test case where we send a weird verb in the request.

proto/connectrpc/conformance/v2/client_compat.proto
smaye81: Either way is fine. Let's stick with `half_close` but maybe just add a bit more detail on what that actually means in the comments?

proto/connectrpc/conformance/v2/service.proto
smaye81: Gotcha, no worries. Just a thought.

proto/connectrpc/conformance/v2/client_compat.proto
jhump: Done. Updated a couple of others that also don't allow negative values (like HTTP status code)

proto/connectrpc/conformance/v2/client_compat.proto
jhump: Doh. I just renamed it and updated the docs 🤦.

Maybe take another look to see if you like the updates?

smaye81: Looks great. Thanks!

smaye81 (approved):

PR #673, opened 2023-11-17
V2: Make error message optional in test case definition

jhump:
This was another change we discussed the other day. This updates the proto and also updates all of the code in the test runner to make it work. So if you were waiting on this to finish the "unimplemented" test, you should be able to proceed now.

This also should make the output when errors and error details don't match a little easier to understand.


internal/app/connectconformance/results.go
```diff
@@ -382,3 +380,47 @@ func checkPayloads(expected, actual []*conformancev2.ConformancePayload) multiEr
...
+	if expected.Code != actual.Code {
+		errs = append(errs, fmt.Errorf("actual error code %d (%s) does not match expected code %d (%s)",
+			actual.Code, connect.Code(actual.Code).String(), expected.Code, connect.Code(expected.Code).String()))
+	}
+	if expected.Message != nil && expected.GetMessage() != actual.GetMessage() {
+		errs = append(errs, fmt.Errorf("actual error message %q does not match expected message %q",
+			actual.GetMessage(), expected.GetMessage()))
+	}
+	if len(expected.Details) != len(actual.Details) {
+		// TODO: Should this be more lenient? Are we okay with a Connect implementation adding extra
```
smaye81, line 405: IMO, this is OK. I think we just want to verify an implementation is able to specify error details. As long as they specify what we instruct them to, they're free to add anything on top of that (at least I think so).

smaye81 (approved):

PR #674, opened 2023-11-17
V2: Looks like it's going to be "v1" after all 😂

jhump:
Renaming the protos back to "v1". So I guess this whole branch (and the prefix on all of the PRs) is a bit mis-named now. Oh well.

This is a mechanical change via `sed` to update all "v2" references to "v1".

And then I did make one manual change: it's the first thing in the diff actually. I changed the connect generated code go to the same place as the base types, since that is consistent with how we do things in our other Go repos.


smaye81 (approved):
lol i forgot about all the PR titles too. oh well 🤷‍♂️

PR #676, opened 2023-11-18
V2 (V1?): Set conformance payload into error details for unary responses

smaye81:
This updates a testing rule that server implementations should set the `ConformancePayload` into error details for unary responses that return an error.


internal/app/connectconformance/results.go
```diff
@@ -413,13 +413,31 @@ func checkError(expected, actual *conformancev1.Error) multiErrors {
 	if len(actual.Details) < length {
 		length = len(actual.Details)
 	}
+	actualPayload := &conformancev1.ConformancePayload{}
+	expectedPayload := &conformancev1.ConformancePayload{}
 	for i := 0; i < length; i++ {
 		// TODO: Should this be more lenient? Are we okay with details getting re-ordered?
 		//       An alternative might be to create a map keyed by type, and for each type
 		//       remove expected messages as they are matched against actual ones.
```
smaye81, line 421: I'm actually wondering whether we care about anything at all other than `ConformancePayload`? From a testing standpoint, we really only care that the server can support error details and the client can read what the server sent. Setting the `ConformancePayload` into error details for unary errors should actually cover that case maybe?

internal/app/connectconformance/testsuites/basic.yaml
```diff
@@ -103,11 +103,6 @@ testCases:
         error:
           code: 13
           message: "server stream failed"
-          details:
```
smaye81, line 106: Removed from all but `grpc-only.yaml` because that file throws the following when details are removed:

```bash
could not write request: client closed stdin
```

Will fix in follow-up PR.

smaye81: Figured out this issue for the follow-up PR, but probably needs a discussion. I'll post it once this one is hashed out.

internal/app/connectconformance/results.go
jhump: We also care about how unprocessable messages in an `Any` are handled, like if the server gets an `Any` it doesn't understand from a remote server and then must include it in the error details. This mainly impacts the Connect protocol, since it tries to parse the `Any` to produce the debug info.

Also, I'm not sure it's `ConformancePayload` we care about. I think it's only `RequestInfo`. The `ConformancePayload` also includes a `data` field, that is only used with real responses (when the response definition indicates the data it should return).

internal/app/connectconformance/testsuites/basic.yaml
jhump: I'm pretty sure that error means that the client program crashed.

internal/app/connectconformance/results.go
```diff
@@ -413,13 +413,31 @@ func checkError(expected, actual *conformancev1.Error) multiErrors {
...
-		if diff := cmp.Diff(expected.Details[i], actual.Details[i], protocmp.Transform()); diff != "" {
-			errs = append(errs, fmt.Errorf("actual error detail #%d does not match expected error detail: - wanted, + got\n%s",
-				i+1, diff))
+		actualDetails := actual.Details[i]
+		expectedDetails := expected.Details[i]
+
+		// If the error details is a conformance payload, then verify equality using the checkPayloads function
+		// Otherwise, just do a straight diff of the two
+		if actualDetails.MessageIs(actualPayload) && expectedDetails.MessageIs(expectedPayload) {
+			// TODO - Do these error conditions need checked? They should prob never fail because of the above check
```
jhump, line 428: I think we can omit this TODO. We definitely want to check this, to make sure that the result we get from the client (possibly by way of the server) is correctly encoded.

But one thing we _do_ want to do is to skip the call to `checkPayloads` if either of the below `UnmarshalTo` fails. If unmarshaling failed, then we don't have a payload to check.

internal/app/connectconformance/test_case_library.go
```diff
@@ -408,6 +409,20 @@ func populateExpectedUnaryResponse(testCase *conformancev1.TestCase) error {
 	case *conformancev1.UnaryResponseDefinition_Error:
 		// If an error was specified, it should be returned in the response
 		expected.Error = respType.Error
+
+		if respType.Error.Details == nil {
```
jhump, line 413: We also want to do this in server responses where the number of response messages is zero.

internal/app/grpcserver/impl.go
```diff
@@ -303,6 +303,18 @@ func parseUnaryResponseDefinition(
 	if def != nil {
 		switch respType := def.Response.(type) {
 		case *v1.UnaryResponseDefinition_Error:
+			// If error details were not provided to be returned, then the server
+			// should set the conformance payload as the error details for unary responses
+			if respType.Error.Details == nil {
```
jhump, line 308: It's often an anti-pattern to compare slices to `nil`. Rarely do you really intend to distinguish between nil and an empty slice. So it's safest to use `len(respType.Error.Details) == 0`.

Also, why do this only when error details is nil? I think we want to add this to the details, even if non-nil.

proto/connectrpc/conformance/v1/service.proto
```diff
@@ -110,6 +110,8 @@ message UnaryResponseDefinition {
     // Response data to send
     bytes response_data = 2;
     // Error to raise instead of response message
+    // If no error details are explicitly provided to return, servers should
+    // still build a ConformancePayload and set it as the details.
```
jhump, line 114: This wording is unclear. It implies to me that the error should always include a `ConformancePayload` (probably the use of the word "still"). But the implementation only sets this payload if the details are empty.

FWIW, I think the server should just always add the request info. If the details are not empty, it can append the request info.

internal/grpcutil/metadata.go
```diff
@@ -28,6 +28,11 @@ func ConvertMetadataToProtoHeader(
 ) []*v1.Header {
 	headerInfo := make([]*v1.Header, 0, len(src))
 	for key, value := range src {
+		// Omit grpc-status-details-bin from the response trailers reported
```
smaye81, line 31: This fixes the `client closed stdin` problem because the problem results from trying to write `grpc-status-details-bin` base64 encoded data to the output (not entirely sure why -- bc of utf-8? buffer length?) to stdout. But is this the right solution? Doing this removes anything grpc specific from the headers returned to the test. But maybe that's fine bc connect has already tried to parse the grpc stuff by now, so as long as we know the error that grpc returned by now we can discard anything we used to glean that information.

We sort of have to do this here bc it needs to be omitted before we send out the response, but thats probably
ok as long as we know the status and details by now.

internal/app/connectconformance/results.go
smaye81: Yeah I actually started out with just `RequestInfo`, but it was easier to do the `ConformancePayload` because the `checkPayloads` function already existed. But, I split that out into a `checkRequestInfo` function and went with `RequestInfo` instead.

internal/app/connectconformance/testsuites/basic.yaml
smaye81: See `metadata.go`. The issue was trying to write `grpc-status-details-bin` to stdout.

smaye81:
@jhump I was actually going to do a follow-up PR that fixed the `client closed stdin` error and also update the server-stream responses to set error detail also, but I just moved it all to this PR since it's easier to get a holistic view of the changes. Lmk what you think.

internal/grpcutil/metadata.go
jhump: This is very confusing. I don't think we should have to do this, and it would be even more confusing if we suggested that other clients-under-test or server-under-test should also do this.

I think we need to identify the real root cause of _why_ this was causing the problem and (hopefully) resolve in a different way.

internal/app/connectconformance/results.go
```diff
@@ -317,65 +317,69 @@ func headerValsToString(vals []string) string {
...
-			if err != nil {
-				errs = append(errs, fmt.Errorf("request #%d: failed to unmarshal expected message: %w", reqNum, err))
-				continue
-			}
-			diff := cmp.Diff(expectedMsg, actualMsg, protocmp.Transform())
-			if diff != "" {
-				errs = append(errs, fmt.Errorf("request #%d: did not survive round-trip: - wanted, + got\n%s", reqNum, diff))
-			}
-		}
+		errs = append(errs, checkRequestInfo(expectedPayload.GetRequestInfo(), actualPayload.GetRequestInfo(), i)...)
```
jhump, line 382: Instead of passing `i`, I think you can check it here:
```go
if i == 0 {
    // Validate headers, timeout, and query params. We only need to do this once, for first response.
    errs = append(errs, checkRequestInfo(expectedPayload.GetRequestInfo(), actualPayload.GetRequestInfo())...)
}
```
That way you can omit the hard-coded zero in the other call site.

internal/grpcutil/metadata.go
jhump: Having said that, I don't think we need to block this PR. Let's add a TODO to investigate further and remove this when the issue is more clearly understood.

jhump (approved):
Let's add a TODO about that grpc-details-bin thing. Otherwise, LGTM

internal/app/connectconformance/results.go
smaye81: The only problem is that value is also used on line 347 if request lengths don't match:

` errs = append(errs, fmt.Errorf("response #%d: expecting %d request messages to be described but instead got %d", respNum+1, len(expected.GetRequests()), len(actual.GetRequests())))`

smaye81: I just removed the number from that `fmt.Errorf` statement. It probably doesn't provide a ton of info when weighed against having to pass the number in as a param.

internal/app/connectconformance/results.go
```diff
@@ -379,7 +377,9 @@ func checkPayloads(expected, actual []*conformancev1.ConformancePayload) multiEr
 			errs = append(errs, fmt.Errorf("response #%d: expecting data %x, got %x", i+1, expectedPayload.Data, actualPayload.Data))
 		}
 
-		errs = append(errs, checkRequestInfo(expectedPayload.GetRequestInfo(), actualPayload.GetRequestInfo(), i)...)
+		if i == 0 { //nolint:nestif
```
jhump, line 380: I just realized that my suggestion changed the functionality here. It's only _part_ of `checkRequestInfo` that we skip when i != 0 -- the headers, timeout, and query params.

For that, I think `checkRequestInfo` should have a `checkHeaders bool` param. This call site can pass `i == 0` for that param, and the other call site can always provide `true`.

PR #682, opened 2023-11-20
Add support for raw requests and responses to reference implementations

jhump:
This still needs testing. It's also big enough, that I'll probably break it up into separate PRs for the test runner (the smallest portion), the reference server (the largest), and the reference client.


jhump:
Breaking this up into a few smaller PRs and adding tests in other other PRs...

PR #686, opened 2023-11-21
V1: Don't quote headers

srikrsna-buf:
Don't quote headers before checking for equality. If we quote them headers that are already `,` seperated will fail the check.

### DIFF
--- a/internal/app/connectconformance/results.go
+++ b/internal/app/connectconformance/results.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io"
 	"sort"
-	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


timostamm:
The [Headers class](https://developer.mozilla.org/en-US/docs/Web/API/Headers) from the fetch API automatically combines values (with the exception of `Set-Cookie`), and we use it connect-es to represent headers and trailers. The old test suite was updated to support it in https://github.com/connectrpc/conformance/pull/469.

It's not a great situation, and I'd prefer better control over actual values sent over the wire. But even if we roll our own alternative, we would still have to go through fetch `Headers` on web browsers, edge runtimes, Next.js, Deno, and Bun - basically everywhere but vanilla Node.js.

jhump:
@timostamm, thanks for the clarification. In that case, it may be better to do a proper semantic check instead of string check, especially since the code this PR is tweaking isn't meant to be a "canoniocal string" function, but was really intended for printing headers to show to users. It was just me being lazy why this is using the same function, opportunistically, to check for equality.

How about something like #687 instead?

PR #687, opened 2023-11-21
V1: Semantic check for header values

jhump:
How about this as an alternative to #686? @srikrsna-buf, @timostamm, @smaye81


timostamm (approved):
Nice and simple, looks great to me.

srikrsna-buf (approved):
Verified with connect node all tests pass for the server!

smaye81 (approved):

PR #690, opened 2023-11-21
V1: Add details to README

smaye81:
This adds some basic details to the README for getting started with the conformance tests.


jhump (approved):
I think we want some sort of `Status` section in here, or maybe even a call-out at the top, too, about the fact that we're still working on an initial v1.0 release, and that it should not be considered stable until after that.

Something similar to [this](https://github.com/bufbuild/prototransform#status).

PR #691, opened 2023-11-21
Helpers for writing raw HTTP body; handle test cases with raw request/response in test runnerf

jhump:
This is a preliminary change that adds a couple of `internal` helpers that the reference client and server will use and to add a little validation to the test runner related to the use of raw requests and responses.

### DIFF
--- a/internal/app/connectconformance/server_runner.go
+++ b/internal/app/connectconformance/server_runner.go
@@ -176,7 +176,7 @@ func runTestCasesForServer(
 				results.assert(name, expectations[resp.TestName], resp.GetResponse())
 			}
 			if isReferenceClient {
-				for _, msg := range resp.Feedback {
+				for _, msg := range resp.GetFeedback() {
 					results.recordSideband(resp.TestName, msg)
 				}
 			}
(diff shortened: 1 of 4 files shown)


internal/app/connectconformance/server_runner.go
```diff
@@ -176,7 +176,7 @@ func runTestCasesForServer(
 				results.assert(name, expectations[resp.TestName], resp.GetResponse())
 			}
 			if isReferenceClient {
-				for _, msg := range resp.Feedback {
+				for _, msg := range resp.GetFeedback() {
```
jhump, line 179: This is an unrelated fix. I encountered a nil-pointer-defereference here at one point while developing this arc of work in #682. Since this PR has other changes in the test runner, seemed okay to drop this here.

internal/headers.go
```diff
@@ -32,6 +32,19 @@ func AddHeaders(
 	}
 }
 
+// AddTrailers adds all header values in src to dest, but
+// it prefixes each header name with http.TrailerPrefix.
```
jhump, line 36: This is how to add trailers from a server handler with `net/http`.

smaye81 (approved):

PR #693, opened 2023-11-22
V1: Add releaser integration

smaye81:
This adds [goreleaser](https://goreleaser.com/) to the repo to enable easy publishing of releases.


.goreleaser.yml
```diff
@@ -0,0 +1,24 @@
...
+    main: ./cmd/connectconformance
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+      - arm64
+    ldflags:
+      - -s -w -X main.buildVersion=v{{.Version}}
```
jhump, line 14: This won't work until we also merge #702. Also, this will need to be tweaked since the `buildVersion` var is not in the `main` package:
```suggestion
      - -s -w -X connectrpc.com/conformance/internal.buildVersion=v{{.Version}}
```

PR #696, opened 2023-11-24
Clarify rules and behavior around response definitions

smaye81:
This tries to clarify the rules around response definitions and how the server should behave in various scenarios where response definition is either not set at all or is partially set with various values.

The basic rules should now be:

TBD


internal/app/grpcclient/impl.go
```diff
@@ -275,30 +284,32 @@ func (i *invoker) bidiStream(
...
 	if err != nil {
 		return nil, err
 	}
-	defer func() {
-		if result != nil {
-			// Set headers and trailers from the stream
-			result.ResponseHeaders = grpcutil.ConvertMetadataToProtoHeader(hdr)
-			result.ResponseTrailers = grpcutil.ConvertMetadataToProtoHeader(stream.Trailer())
-		}
-	}()
```
smaye81, line 301: This had to be rearranged a bit because in instances where we returned early (i.e. line 281 above), the `defer` wasn't defined yet, so headers and trailers were not getting set.

internal/app/connectconformance/results.go
```diff
@@ -110,8 +110,8 @@ func (r *testResults) failed(testCase string, err *conformancev1.ClientErrorResu
 func (r *testResults) assert(testCase string, expected, actual *conformancev1.ClientResponseResult) {
 	var errs multiErrors
 
-	if len(expected.Payloads) == 0 && expected.Error != nil {
```
smaye81, line 113: I changed this because gRPC-Web seems to lump the headers and trailers together even if an immediate error isn't thrown. So, all gRPC-Web tests were failing on trailers-only responses that didn't return an error.

You can see this behavior if you send a server stream request that expects no error and no responses via gRPC-Web. The desired response headers and trailers are all combined into one `responseHeaders` list. gRPC, however, will split them into headers and trailers accordingly.

jhump: Hmmm, might this actually be a _bug_ in the grpc-web implementation then? Maybe I need to re-read this to make sure I understand the behavior and whether we actually want to allow this.

jhump: Let's add a TODO that this check for trailers-only should really only apply to gRPC and gRPC-Web protocols. And the relaxed check, that allows this with no error, should be only for gRPC-Web.

PR #711, opened 2023-11-27
Clarify rules around response definitions

smaye81:
This tries to clarify the rules around response definitions and how the server should behave in various scenarios where response definition is either not set at all or is partially set with various values. To complement this, it also adds tests for each RPC type that have partially-set response definitions as well as no response definition. Numerous areas in the servers had to be fixed to check for a `nil` response definition and to act accordingly. 

The protos have been updated to reflect the rules now and should (hopefully) be clearer and more thorough. The basic rules are:

* for unary responses, if no response definition is set, then return the request info and an empty response
* for server responses, if no response definition is set (or is set, but no responses are desired), do not send anything on the stream and either return without error or return an error if defined.

The bidi flow has changed slightly. The basic premise for full and half duplex is 'if you have no responses to send, check to see if an error was provided. If so, throw that. Otherwise, just do nothing and return without error'. As a result, it is now possible for a full duplex call to have less responses desired than requests sent. In these cases, the flow should follow the above premise.

Also, the `connect-only` and `grpc-only` test files have been merged into `basic` since those should be covered the same for all protocols now.

Note: this replaces #696 since there were way too many merge conflicts with the merge of v1 into main. This should be identical to PR (just based off of main and with no merge conflicts).


internal/app/grpcclient/impl.go
```diff
@@ -104,8 +110,9 @@ func (i *invoker) unary(
 	if err != nil {
 		// If an error was returned, convert it to a gRPC error
 		protoErr = grpcutil.ConvertGrpcToProtoError(err)
-	} else {
-		// If the call was successful, get the returned payloads
+	} else if resp.Payload != nil {
```
jhump, line 113: You can change this back to `} else {` since it is not possible for payload to be nil if err is nil. One or the other will be non-nil since a unary RPC _must_ either return a response or an error.

internal/app/grpcclient/impl.go
```diff
@@ -279,30 +286,31 @@ func (i *invoker) bidiStream(
...
-		result.Error = grpcutil.ConvertGrpcToProtoError(err)
-		return result, nil
+		return nil, err
 	}
 
 	// Once the send side is closed, header metadata is ready to be read
-	hdr, err := stream.Header()
+	hdr, err = stream.Header()
 	if err != nil {
 		return nil, err
```
jhump, line 306: We probably want to return `protoErr` here, and above on line 300, if it's not nil. Generally, `stream.Receive` is the proper way to retrieve the RPC error. I'm not certain if it's returned by `CloseSend` or `Header`.

internal/app/referenceclient/impl.go
```diff
@@ -111,11 +111,13 @@ func (i *invoker) unary(
 		headers = internal.ConvertToProtoHeader(connectErr.Meta())
 		protoErr = internal.ConvertConnectToProtoError(connectErr)
 	} else {
-		// If the call was successful, get the returned payloads
-		// and the headers and trailers
-		payloads = append(payloads, resp.Msg.Payload)
+		// If the call was successful, get the headers and trailers
 		headers = internal.ConvertToProtoHeader(resp.Header())
 		trailers = internal.ConvertToProtoHeader(resp.Trailer())
+		// If there's a payload, add that the response also
```
jhump, line 117:
```suggestion
		// If there's a payload, add that to the response also
```

internal/app/referenceserver/impl.go
```diff
@@ -123,53 +134,56 @@ func (s *conformanceServer) ServerStream(
...
+			// Only set the request info if this is the first response being sent back
+			// because for server streams, nothing in the request info will change
+			// after the first response.
+			if respNum == 0 {
+				resp.Payload.RequestInfo = createRequestInfo(ctx, req.Header(), []*anypb.Any{msgAsAny})
+			}
 
-		if err := stream.Send(resp); err != nil {
-			return connect.NewError(connect.CodeInternal, fmt.Errorf("error sending on stream: %w", err))
+			time.Sleep((time.Duration(responseDefinition.ResponseDelayMs) * time.Millisecond))
```
jhump, line 164: nit: I didn't catch this in the original review, but this would be more readable if you factor out the sleep duration into a variable, above outside the loop.

jhump (approved):
Cool. Just a couple of little things. LGTM

internal/app/referenceserver/impl.go
smaye81: Sounds good. Should I do the same for unary and client streaming also? Feels like I should for consistency but they aren't in a loop and are sort of a one-shot deal. I'd probably be declaring the variable directly above the `Sleep` call.

internal/app/grpcclient/impl.go
smaye81: Just to clarify, you mean something like this (and same thing for `CloseSend`)?

```
	hdr, err = stream.Header()
	if err != nil {
		result.Error = grpcutil.ConvertGrpcToProtoError(err)
		return result, nil
	}
```

One error where I was concerned is that the return of `protoErr` is after these calls, which means something could've failed above in the send loop and got set to `protoErr` but we're returning here and losing whatever that error was (unless `CloseSend` or `Header` actually returns us that error again as you alluded to)

jhump: No I meant more like:
```go
	hdr, err = stream.Header()
	if err != nil && protoErr == nil {
		protoErr = grpcutil.ConvertGrpcToProtoError(err)
	}
```

internal/app/referenceserver/impl.go
jhump: Sure, that would probably help readability in the other places, too.

PR #715, opened 2023-11-28
Add support for specifying timeouts

smaye81:
This adds support for specifying timeouts and adds unit tests for validating context deadlines. In doing so, it fixes the expectations so that the test library validates servers are returning the received `TimeoutMs`.


proto/connectrpc/conformance/v1/service.proto
```diff
@@ -140,6 +144,9 @@ message UnaryResponseDefinition {
   // Response trailers to send - together with the error if present
   repeated Header response_trailers = 4;
 
+  // Wait this many milliseconds before sending a response message
+  uint32 response_delay_ms = 6;
```
smaye81, line 148: Realized there was no way to specify a response delay for unary responses.

internal/app/connectconformance/testsuites/basic.yaml
```diff
@@ -73,6 +73,21 @@ testCases:
         responseTrailers:
         - name: x-custom-trailer
           value: ["bing","quux"]
+- request:
```
smaye81, line 76: Bidi's seem to be failing timeouts. Investigating in a separate PR.

jhump (approved):

PR #717, opened 2023-11-29
Add ability to specify a test file from command line

smaye81:
This adds the ability to specify a test file via the command line. The given path must be a file in YAML format, otherwise an error is returned.

A scenario for usage is when debugging tests and isolating ones that are failing. The tests can be moved to their own file and also further isolated using the `relevantXXX` directives so the tests only run for those specified configs.

Example usage:

```bash
go run cmd/connectconformance/main.go --mode client .tmp/bin/referenceclient --test-file ./internal/app/connectconformance/testsuites/bidi.yaml
```


smaye81:
@jhump put up this PR because this functionality was invaluable while trying to debug some of the failing timeout stuff. It's super helpful to be able to isolate tests without having to run the whole suite.

jhump (approved):

PR #718, opened 2023-11-30
Fix some error-handling issues in the reference client's server stream impl

jhump:
There were two cases where the `serverStream` method was just doing `return nil, err`, when it really should have been converting the error to a `*conformancev1.Error` and storing it in a non-nil response. That was causing server-stream timeout error to frequently be interpreted as a test case failure, even though the error was a "deadline exceeded" error, and that's also what the test case was expecting.

### DIFF
--- a/internal/app/connectconformance/server_runner.go
+++ b/internal/app/connectconformance/server_runner.go
@@ -175,8 +175,8 @@ func runTestCasesForServer(
 			default:
 				results.assert(name, expectations[resp.TestName], resp.GetResponse())
 			}
-			if isReferenceClient {
-				for _, msg := range resp.GetFeedback() {
+			if isReferenceClient && resp != nil {
+				for _, msg := range resp.Feedback {
 					results.recordSideband(resp.TestName, msg)
 				}
 			}
--- a/internal/app/referenceclient/impl.go
+++ b/internal/app/referenceclient/impl.go
@@ -153,7 +153,16 @@ func (i *invoker) serverStream(
 
 	stream, err := i.client.ServerStream(ctx, request)
 	if err != nil {
-		return nil, err
+		// If an error was returned, first convert it to a Connect error
+		// so that we can get the headers from the Meta property. Then,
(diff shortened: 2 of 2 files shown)


internal/app/referenceclient/impl.go
```diff
@@ -267,11 +278,6 @@ func (i *invoker) bidiStream(
 
 	var protoErr *v1.Error
 	for _, msg := range req.RequestMessages {
-		if err := ctx.Err(); err != nil {
-			// If an error was returned, convert it to a proto Error
-			protoErr = internal.ConvertErrorToProtoError(err)
-			break
-		}
```
jhump, line 274: This is unnecessary. If the context is cancelled, it will get noticed and caught elsewhere. Also, we don't need to do anything special since at this point the client has cancelled the operation and won't be looking at any response.

internal/app/referenceclient/impl.go
```diff
@@ -246,7 +257,7 @@ func (i *invoker) clientStream(
 func (i *invoker) bidiStream(
 	ctx context.Context,
 	req *v1.ClientCompatRequest,
-) (result *v1.ClientResponseResult, retErr error) {
+) (result *v1.ClientResponseResult, _ error) {
```
jhump, line 260: This was unused.

internal/app/connectconformance/server_runner.go
```diff
@@ -175,8 +175,8 @@ func runTestCasesForServer(
 			default:
 				results.assert(name, expectations[resp.TestName], resp.GetResponse())
 			}
-			if isReferenceClient {
-				for _, msg := range resp.GetFeedback() {
+			if isReferenceClient && resp != nil {
+				for _, msg := range resp.Feedback {
```
jhump, line 179: Unrelated tweak: we end up using `resp.TestName` below (without accessor), so it seemed more obviously correct to hoist the nil check up. And then we can safely use direct field access in this block.

smaye81 (approved):

PR #719, opened 2023-11-30
Support grpc-web with the grpc-go server

jhump:
This allows us to test the grpc-web protocol against the grpc-go implementation. This will be utilized for all clients that support grpc-web.


Makefile
```diff
@@ -95,13 +95,14 @@ runconformance: runservertests runclienttests
 
 .PHONY: runservertests
 runservertests: $(BIN)/connectconformance $(BIN)/referenceserver $(BIN)/grpcserver
-	$(BIN)/connectconformance --conf ./testdata/reference-impls-config.yaml --mode server $(BIN)/referenceserver
-	$(BIN)/connectconformance --conf ./testdata/grpc-impls-config.yaml --mode server $(BIN)/grpcserver
+	$(BIN)/connectconformance -v --conf ./testdata/reference-impls-config.yaml --mode server -- $(BIN)/referenceserver
+	$(BIN)/connectconformance -v --conf ./testdata/grpc-impls-config.yaml --mode server -- $(BIN)/grpcserver
+	$(BIN)/connectconformance -v --conf ./testdata/grpc-web-server-impl-config.yaml --mode server -- $(BIN)/grpcserver
```
jhump, line 100: Since the client and server have different capabilities, I split the config into two. One represents what the client supports, which the server also supports (`grpc-impls-config.yaml`), and this new file (`grpc-web-server-impl-config.yaml`) is just the incremental cases on top of that now that the server supports the gRPC-Web protocol.

internal/app/grpcserver/server.go
```diff
@@ -96,11 +117,28 @@ func Run(ctx context.Context, args []string, inReader io.ReadCloser, outWriter i
...
-func createServer() (*grpc.Server, error) { //nolint:unparam
-	server := grpc.NewServer(
-		grpc.UnaryInterceptor(serverNameUnaryInterceptor),
-		grpc.StreamInterceptor(serverNameStreamInterceptor),
-	)
-	v1.RegisterConformanceServiceServer(server, NewConformanceServiceServer())
-	return server, nil
+func runGRPCWebServer(ctx context.Context, server *grpc.Server, listener net.Listener) error {
+	httpServer := http.Server{
+		Handler:           h2c.NewHandler(grpcweb.WrapServer(server), &http2.Server{}),
```
jhump, line 122: This `grpcweb.WrapServer` call is the special-sauce that makes this work with grpc-web.

smaye81 (approved):

PR #720, opened 2023-11-30
Speed up conformance runs with parallelism

jhump:
This introduces two new flags to the `connectconformance` runner:
1. `max-servers`: the number of concurrent server processes to start. Test cases for all of them will be written to the client's stdin in parallel (thread-safe, but interleaving test cases for various servers). Defaults to 4.
2. `parallel`: the number of RPCs to issue in parallel when `--mode` is `server` (so this only controls the reference client). In `--mode` of `client`, it's up to the client-under-test to implement parallelism. Defaults to 4 * GOMAXPROC (and GOMAXPROC defaults to the number of CPU cores detected).

This speeds up conformance runs on my MBP from about 3 minutes down to less than 15 seconds,.

### DIFF
--- a/cmd/connectconformance/main.go
+++ b/cmd/connectconformance/main.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"connectrpc.com/conformance/internal"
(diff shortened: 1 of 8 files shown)


jhump:
@smaye81, ping

smaye81:
Nice! Tested this on my machine too and it was _much_ faster.

smaye81 (approved):

PR #721, opened 2023-12-03
Add tests for Connect GET operations

smaye81:
This adds tests to the conformance suite for Connect GET operations via a new endpoint named `IdempotentUnary`. This endpoint is annotated as `NO_SIDE_EFFECTS` and all clients and servers should use HTTP GET to communicate.


internal/codec.go
```diff
@@ -110,6 +112,22 @@ func (j *jsonEncoder) Encode(msg proto.Message) error {
 	return nil
 }
 
+// This function is lifted from connect-go since this is the logic it uses
```
smaye81, line 115: It doesn't feel right to duplicate this logic here either. I think ideally it would be nice if the same logic that the Go implementation was using to encode was made available, but that doesn't feel right either. Let me know if there's a better way to handle this.

internal/app/connectconformance/test_case_library.go
```diff
@@ -410,18 +409,35 @@ func hasRawResponse(reqs []*anypb.Any) bool {
 	return false
 }
 
+// returns the response definition from an Any representing a unary request.
```
jhump, line 412: nit: Go doc convention is to always start with the element name.
```suggestion
// getUnaryResponseDefinition returns the response definition from an Any representing a unary request.
```

internal/app/connectconformance/test_case_library.go
```diff
@@ -430,9 +446,44 @@ func populateExpectedUnaryResponse(testCase *conformancev1.TestCase) error {
...
+				{
+					Name:  "message",
+					Value: []string{value},
+				},
+				{
+					Name:  "encoding",
+					Value: []string{encoding},
+				},
+				// TODO: How should we verify the connect version parameter?
+				// Match the name "connect" and a value that is v + some number?
```
jhump, line 478: I must always be "v1". We don't need to worry about making this more lenient until we actually support some newer version of the protocol.

internal/app/connectconformance/test_case_library.go
```diff
@@ -410,18 +409,35 @@ func hasRawResponse(reqs []*anypb.Any) bool {
...
 // populates the expected response for a unary test case.
 func populateExpectedUnaryResponse(testCase *conformancev1.TestCase) error {
 	req := testCase.Request.RequestMessages[0]
-	// First, find the response definition that the client instructed the server to return
-	concreteReq, err := req.UnmarshalNew()
-	if err != nil {
-		return err
+	var msg unaryResponseDefiner
+
+	switch req.TypeUrl {
```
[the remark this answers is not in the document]
jhump: I am struggling to understand this change. Why do we need to know the actual request type to marshal it? Marshalling only needs a `proto.Message`, not any concrete type. Also, the `msg` var is still declared above with the type `unaryResponseDefiner`, so when it's marshalled below, we do not know the actual request type. We only know it's a `unaryResponseDefiner`.

internal/app/connectconformance/test_case_library_test.go
```diff
@@ -901,6 +901,44 @@ func TestPopulateExpectedResponse(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	idempotentUnaryErrorJSONReqInfo, err := anypb.New(&conformancev1.ConformancePayload_RequestInfo{
```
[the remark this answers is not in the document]
jhump: The better way is to use JSON here. The struct literal syntax is much more verbose (especially since it requires lots of error-checking code to construct `Any` messages).

See `TestExpandRequestData` for an example.

internal/app/connectconformance/test_case_library_test.go
```diff
@@ -901,6 +901,44 @@ func TestPopulateExpectedResponse(t *testing.T) {
...
+
+	idempotentUnaryErrorProtoReqInfo, err := anypb.New(&conformancev1.ConformancePayload_RequestInfo{
+		RequestHeaders: requestHeaders,
+		Requests:       []*anypb.Any{unaryErrorReq},
+		ConnectGetInfo: &conformancev1.ConformancePayload_ConnectGetInfo{
+			QueryParams: []*conformancev1.Header{
+				{
+					Name: "message",
+					// unaryErrorReq marshaled and base64 encoded
+					Value: []string{"Cq4BChkKCWZvb0hlYWRlchIMZm9vSGVhZGVyVmFsCikKCWJhckhlYWRlchINYmFySGVhZGVyVmFsMRINYmFySGVhZGVyVmFsMiIbCgpmb29UcmFpbGVyEg1mb29UcmFpbGVyVmFsIiwKCmJhclRyYWlsZXISDmJhclRyYWlsZXJWYWwxEg5iYXJUcmFpbGVyVmFsMhobCAgSF2FsbCByZXNvdXJjZXMgZXhoYXVzdGVk"},
```
jhump, line 931: This and the JSON-serialized message above are likely best done by creating a message literal and then serializing it. Otherise, it's not clear how one updates the test case if something needs to be tweaked (like a field here or there).

jhump: If most of the definition is defined using JSON (like suggested above), you could leave out parts like that from the JSON and also supply the message instances, and then serialize them and store them inside the reuqest/response definition inside the loop that processes the test cases.

internal/app/referenceclient/impl.go
```diff
@@ -136,6 +145,52 @@ func (i *invoker) unary(
 	}, nil
 }
 
+func (i *invoker) idempotentUnary(
```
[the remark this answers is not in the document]
jhump: If the only difference is the request and response type, you could make a generic function that all three invoke.

Something with a signature like so (note that it must be a function, not a method, because methods in Go cannot be generic):
```go
// These first two types are generic constraints.
type pointerMsg[T any] interface {
	*T
	proto.Message
}

type payloadContainer[T any] interface {
	pointerMsg[T]
	GetPayload() *v1.ConformancePayload
}

func invokeUnary[ReqMsg pointerMsg[ReqT], RespMsg payloadContainer[RespT], ReqT, RespT any](
	ctx context.Context,
	req *v1.ClientCompatRequest,
	method func(context.Context, *connect.Request[ReqT]) (*connect.Response[RespT], error),
) (*v1.ClientResponseResult, error) {
	// use generic type to construct request
	ur := ReqMsg(new(ReqT))
	// ...

	// and also to invoke RPC
	connResp, err := method(ctx, connect.NewRequest(ur))
	if err != nil {
		return nil, err
	}
	resp := RespMsg(connResp.Msg)
	// ...
} 

// You can call the above like so to pass the particular RPC endpoint
// as a generic function parameter.
resp, err := invokeUnary(ctx, req, i.client.IdempotentUnary)
```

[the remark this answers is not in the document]
jhump: You could always add a TODO to consolidate in the future. For now, I think it's okay that `idempotent` is different, so we really only replicate all of the logic 2x.

internal/app/referenceserver/checks.go
```diff
@@ -145,12 +146,15 @@ func checkCodec(expected v1.Codec, req *http.Request, feedback *feedbackWriter)
...
 	}
-	contentType, hasContentType := getHeader(req.Header, "content-type", feedback)
+	contentType, _ := getHeader(req.Header, "content-type", feedback)
 	var actual string
 	switch {
 	case req.Method == http.MethodGet:
-		if hasContentType {
-			feedback.Printf("content-type header should not appear with method GET")
+		// Servers should test for an empty request body by attempting a read.
+		// If no body is present, it should return an immediate EOF.
```
jhump, line 154: I think we want to keep the `hasContentType` var above and then have something like so here:
```
// TODO: Require that no content-type be present. Must wait for
//       https://github.com/connectrpc/connect-go/pull/644 to
//       be release and incorporated into the reference client.
_ = hasContentType
```

internal/app/referenceserver/impl.go
```diff
@@ -55,14 +56,10 @@ func (s *conformanceServer) Unary(
...
 		req.Header(),
+		nil,
 		[]*anypb.Any{msgAsAny},
 	)
-	// If connectErr is not nil, then the response definition must not be nil.
-	// Read the headers and trailers from the definition and add them to connectErr
-	// and return
 	if connectErr != nil {
-		internal.AddHeaders(req.Msg.ResponseDefinition.GetResponseHeaders(), connectErr.Meta())
-		internal.AddHeaders(req.Msg.ResponseDefinition.GetResponseTrailers(), connectErr.Meta())
```
jhump, line 65: Why remove these?

internal/app/referenceserver/impl.go
```diff
@@ -82,6 +79,42 @@ func (s *conformanceServer) Unary(
 	return resp, nil
 }
 
+func (s *conformanceServer) IdempotentUnary(
+	ctx context.Context,
+	req *connect.Request[v1.IdempotentUnaryRequest],
+) (*connect.Response[v1.IdempotentUnaryResponse], error) {
+	if req.HTTPMethod() != http.MethodGet {
+		return nil, errors.New("idempotent unary method should have been called via HTTP GET")
+	}
```
jhump, line 88: FWIW, this should already be checked by the reference server in `checks.go`.

This should be pretty straight-forward to share code with `Unary`, right? Maybe at least a TODO?

internal/app/referenceserver/impl.go
```diff
@@ -55,14 +56,10 @@ func (s *conformanceServer) Unary(
 		ctx,
 		req.Msg.ResponseDefinition,
 		req.Header(),
+		nil,
```
jhump, line 59: Let's pass `req.Peer().Query` from everywhere, instead of hard-coding to nil. That way if query string params are inadvertently provided for POST endpoints, the reference server can detect that.

internal/codec.go
jhump: I don't see any other way right now, so this seems fine.

It's a pet peave of mine that the connect-go library doesn't expose any of this (codec impls) for re-use. Ideally, it would just make its codec and compression impls public so you can wrap/re-use. 🤷

internal/headers.go
```diff
@@ -45,9 +46,20 @@ func AddTrailers(
...
+func ConvertHTTPHeaderToProtoHeader(src http.Header) []*v1.Header {
+	return convertToProtoHeader(src)
+}
+
+// ConvertQueryParamToProtoHeader converts query parameters to a slice of proto Headers.
+func ConvertQueryParamToProtoHeader(
+	src url.Values,
+) []*v1.Header {
+	return convertToProtoHeader(src)
+}
```
jhump, line 59: Not sure I see the value of these two specialized methods. Why not just change the type of the one, exported `ConvertToProtoHeader` to accept `map[string][]string` instead of `http.Header` and then document in the comments that it can accept either `http.Header` or `url.Values`.

internal/app/connectconformance/test_case_library_test.go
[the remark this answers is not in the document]
smaye81: I created a TODO for this to refactor this if that's cool. I think I'd rather do that as part of its own PR.

Won't converting to JSON still lead to a bunch of repeated text since I can't (easily) reuse strings for say the headers/trailers, etc.? Not sure if it would be a wash with verbosity.

internal/app/referenceclient/impl.go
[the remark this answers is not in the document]
smaye81: Yep. Done.

internal/app/referenceserver/impl.go
smaye81: I didn't actually remove these, I just moved them to be added in `parseUnaryDefinition`.

internal/app/connectconformance/test_case_library.go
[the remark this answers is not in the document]
smaye81: Yep, good call. All I really needed was to add `proto.Message` to the `unaryResponseDefiner` and that took care of it. Thanks.

internal/app/referenceserver/impl.go
smaye81: Yeah, but it still involves the same issue as the client -- the different request/response type. I added a TODO to handle this the same way.

internal/headers.go
smaye81: Ah ok. I figured we'd want the type safety of passing either `url.Value` or `http.Header`, but using the map as the public API is cool with me.

internal/app/referenceserver/impl.go
jhump: Yeah, my bad. I did see that and thought I removed this draft comment.

internal/app/connectconformance/test_case_library_test.go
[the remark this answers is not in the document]
jhump: > I created a TODO for this to refactor this if that's cool. 

Most definitely. 👍

internal/app/referenceserver/impl.go
jhump: > the different request/response type

This case is much simpler though: you just don't put that part in the shared function. Have the shared function accept the parts of the request (`requestData []byte` and `*UnaryResponseDefinition`) and return the `*ConformancePayload`. Then the only thing these two RPC methods do is extract those two parameters, call the shared function, and then package the payload in the right response type.

> I added a TODO to handle this the same way.

Cool, that's fine.

PR #726, opened 2023-12-07
Add tests for cancellation

smaye81:
This adds tests for cancellation logic based on the `cancel_timing` oneof in the test request.


internal/app/grpcclient/impl.go
```diff
@@ -164,8 +174,19 @@ func (i *invoker) serverStream(
...
+	}
+	// If the cancel timing specifies after 0 responses, then cancel before
+	// receiving anything
+	if timing.AfterNumResponses == 0 {
+		ctxCancel()
+	}
+	totalRcvd := 0
 	for {
 		msg, err := stream.Recv()
+		totalRcvd++
```
smaye81, line 189: For some reason, if `AfterNumResponses == 0` and the context was cancelled, the `stream.Recv` call above returns a message and a `nil` error on the first iteration.

internal/app/connectconformance/testsuites/cancellation.yaml
```diff
@@ -0,0 +1,100 @@
+name: Cancellation
+testCases:
+# TODO - The server stream tests are failing with grpc due to issues with
```
smaye81, line 3: As we discussed offline, this is because of the `stream.Header` call in the gRPC client implementation. That call blocks until header metadata is sent, but the server stream endpoint sends the header information with the first response. So, the `stream.Header` call blocks until a response is sent and _then_ the context is cancelled. So right now, the expectations for responses on server streams won't work.

This can be fixed with a change on the reference server and the gRPC server:

Reference Server
- should call `stream.Send(nil)` after setting the headers.

gRPC Server
- should call `stream.SendHeader(headerMD)` (instead of `SetHeader`)

However, doing the above causes a "protocol error: incomplete envelope: context deadline exceeded" error to be thrown by Connect-Go on the server stream timeout tests.

internal/cancellation.go
```diff
@@ -0,0 +1,68 @@
...
+
+// WrapContext wraps the current context. The resulting context and cancel
+// function will be dependent on whether the given context has a deadline.
+func WrapContext(ctx context.Context) (context.Context, context.CancelFunc) {
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		return context.WithCancel(ctx)
+	}
+	return context.WithDeadline(ctx, deadline)
+}
```
jhump, line 68: This is not necessary. You can unconditionally use `context.WithCancel` to get a cancel function. If it has a deadline, that's fine -- it will be propagated to the new child context created by `context.WithCancel`.

internal/app/grpcclient/impl.go
```diff
@@ -132,9 +132,12 @@ func (i *invoker) unary(
 }
 
 func (i *invoker) serverStream(
-	ctx context.Context,
+	c context.Context,
 	ccr *v1.ClientCompatRequest,
 ) (result *v1.ClientResponseResult, retErr error) {
+	ctx, ctxCancel := internal.WrapContext(c)
```
jhump, line 138: Here and throughout.
```suggestion
	ctx, cancel := context.WithCancel(ctx)
```

internal/app/connectconformance/testsuites/bidi-cancellation.yaml
```diff
@@ -0,0 +1,226 @@
+name: Bidi Cancellation
+relevantHttpVersions:
+  - HTTP_VERSION_2
+  - HTTP_VERSION_3
```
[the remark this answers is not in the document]
smaye81: Nice. I combined `bidi.yaml` into `basic.yaml` then since we didn't need to specify HTTP versions.

jhump (approved):

PR #732, opened 2023-12-11
Run all non-TLS-specific test cases through TLS server

jhump:
This effectively adds a lot more test cases for TLS servers. Previously, only test cases that _required_ TLS were run against TLS servers. Now, all test cases are run against TLS servers.

We still only run test cases that require client certs when using the client cert configuration. But for TLS servers without client certs, we run 'em all.

I ran into some gross issues with HTTP/3 with this branch, which is where the changes to the reference client and server come from.


jhump:
cc @srikrsna-buf

smaye81 (approved):

PR #734, opened 2023-12-12
Fix cancellation and timeout failures

smaye81:
This makes various improvements to fix the failing cancellation and timeout tests:

* Removes `ctx.Err()` checks in the clients since this can be an ambiguous error on the client and could potentially be unrelated to the error we received from the RPC framework.
* Adds logic to immediately send response headers on server and bidi streams so that clients can be unblocked reading this data. Previously this was causing clients to block on reading this data and was preventing them from cancelling at the intended time.
* Adds docs to the protos specifying the above.
* Adds cancellation logic to gRPC `BidiStream`
* Configures some tests with appropriate delays so that the cancellation is properly communicated.

All of the tests in `bidi-cancellation.yaml` and `bidi-timeouts.yaml` have been folded into the `basic.yaml` suite of tests since these apply to and pass all permutations.


internal/app/connectconformance/testsuites/basic.yaml
```diff
@@ -333,6 +436,149 @@ testCases:
         - name: x-custom-trailer
           value: ["bing","quux"]
       fullDuplex: true
+- request:
```
jhump, line 439: I think it would be nice to keep the timeouts and cancellations in separate files and have "basic" just be basic happy-path RPCs.

Also, I think for the cancellations file, we'd want `mode: TEST_MODE_CLIENT` at the top, since there's not much use in running these test cases against a server under test, since we can't actually assert anything about how the server behaves in the face of cancellation -- all we can really see/verify is that the client-side RPC framework fails the call with a "cancelled" code.

internal/app/grpcserver/impl.go
```diff
@@ -286,6 +288,9 @@ func (c *conformanceServiceServer) BidiStream(
 	// where the requested responses are greater than the total requests.
 	if responseDefinition != nil { //nolint:nestif
 		for ; respNum < len(responseDefinition.ResponseData); respNum++ {
+			if err := stream.Context().Err(); err != nil {
+				return err
+			}
```
jhump, line 293: While this seems useful, to let the server abort on cancellation/timeout, it wasn't necessary to get any of the test cases to pass was it?

FWIW, this kind of thing isn't really needed because the very next stream interaction (`stream.Send` below) should return an error if the RPC has been cancelled.

jhump (approved):
Though bidi stuff doesn't need to be broken out into separate files from the other stream types, I think breaking out timeout and cancel cases into their own test suite files was better than putting everything into a single suite.

Otherwise, LGTM.

PR #735, opened 2023-12-12
Add flags to test runner allowing user to supply TLS cert+key, port, and bind address for in-process reference server(s)

jhump:
In certain environments, it may not be possible to configure trust roots and use self-signed certificates. Also, it may be necessary to know/configure the listen port so that the reference server can be properly exported for use by an external client. For these reasons, this adds flags to the test runner that allow the user to supply a TLS cert and key and to configure the port and bind address on which reference server(s) listen. Without these, the servers will generate a self-signed cert+key and will listen on "127.0.0.1:0".

This caught an issue in the HTTP/3 server: it's `CloseGracefully` method **is not actually implemented!**. [WTF?](https://github.com/quic-go/quic-go/blob/a3603549eebd4e22fd8ea4d4fe0c9011bc08daa4/http3/server.go#L666-L669) 😠

This was an issue because when all server process are using the same port, this meant that the HTTP/3 process never closed and never stopped its listener, so the next process started was unable to re-use the port.

In order to troubleshoot that, I also added code to the server runner so that it will propagate lines to stderr that don't appear to be sideband feedback about the client under test and to the in-process starter to log any errors to stderr (they were previously being swallowed 😢).

While in here, I also updated the grpc-go server so that it will report an error if it gets a configuration that it can't handler (instead of proceeding to start-up an incorrectly configured server).


jhump:
cc @srikrsna-buf

smaye81 (approved):

PR #739, opened 2023-12-14
Add tests for message size

smaye81:
This adds tests for message size for both client mode and server mode.

### DIFF
--- a/internal/app/connectconformance/testsuites/client_message_size.yaml
+++ b/internal/app/connectconformance/testsuites/client_message_size.yaml
@@ -1,7 +1,6 @@
 name: Client Message Size
 mode: TEST_MODE_CLIENT
 reliesOnMessageReceiveLimit: true
-# TODO - The size limit is currently only calculated against the proto size
 relevantCodecs:
   - CODEC_PROTO
 testCases:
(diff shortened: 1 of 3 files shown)


internal/app/connectconformance/testsuites/client_message_size.yaml
```diff
@@ -0,0 +1,251 @@
+name: Client Message Size
+mode: TEST_MODE_CLIENT
+reliesOnMessageReceiveLimit: true
+# TODO - The size limit is currently only calculated against the proto size
+relevantCodecs:
+  - CODEC_PROTO
```
smaye81, line 6: @jhump the request expansion logic seems to just use the proto codec for size calculations. I'm not sure if we should also test at least the JSON codec, but my hunch is that testing with proto only is probably fine?

jhump: Ah, yes. Sorry I forgot to mention it to you. It may be worth adding a check when we load the test case library to complain if it sees a request that uses the "expand request" stuff but with a codec _other than_ proto.

jhump (approved):
LGTM!

PR #740, opened 2023-12-14
Support sending raw HTTP response in reference server

jhump:
This one's a little more confusing the client side because of how we to hook into the HTTP framework's response writer. We basically override the response writer to _maybe_ ignore all calls to it (made by the handler to send its "normal" HTTP response) and, at the end, instead send the raw response. Because it waits until the end to send the raw response, we can't test full-duplex bidi stuff with it. But it does work fine with half-duplex bidi operations.

The "maybe" when ignoring calls is whether there is a raw response or not. So there's a connect interceptor that examines the request as the first thing and, if it contains a raw response, stashes it away so the overridden response writer knows about it. This "stashing away" uses a placeholder address in a context value.

### DIFF
--- a/internal/app/referenceclient/raw_request_test.go
+++ b/internal/app/referenceclient/raw_request_test.go
@@ -54,6 +54,26 @@ func TestRawRequestSender(t *testing.T) {
 	err = anypb.MarshalFrom(msgPayload, val, proto.MarshalOptions{})
 	require.NoError(t, err)
 
+	var requests sync.Map // map[string]chan *http.Request
+	svr := httptest.NewServer(http.HandlerFunc(func(respWriter http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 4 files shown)


internal/app/referenceclient/raw_request_test.go
```diff
@@ -54,6 +54,26 @@ func TestRawRequestSender(t *testing.T) {
 	err = anypb.MarshalFrom(msgPayload, val, proto.MarshalOptions{})
 	require.NoError(t, err)
 
+	var requests sync.Map // map[string]chan *http.Request
```
jhump, line 57: The changes in this file are admittedly unrelated. I was just moving stuff around to make it a little more consistent with the other tests (and maybe a little more readable?).

smaye81 (approved):

PR #742, opened 2023-12-14
Refactor `TestPopulateExpectedResponse` tests

smaye81:
This PR attempts to make some improvements to the `TestPopulateExpectedResponse` test cases. Currently, they are very hard to follow and add to due to how messages are constructed ahead of time and then set in the `testCase` list. It is confusing to look at an individual test and figure out what's going on without bouncing around to the various definitions.

The reason this was originally coded this way is because a vast majority of these messages needed converted to an `Any` and `anypb.New` returns a message and an error, so that error needed verified. To alleviate this and help remove these pre-constructed messages, a new function has been created: `asAnySlice`, which takes a variable amount of messages and converts them to a slice of `Any` messages. It also verifies the error itself so that the callsites can now just accept a single value.

Note that this PR does not add any additional tests and it admittedly does not shrink the size of the `TestPopulateExpectedResponse` test. But, each test case should now be self-contained and easier to grok by just looking at the `testCase.request` and `testCase.expected` values without having to bounce all over the place. It should also be much easier to add new tests without having to pre-construct a multitude of message definitions.

### DIFF
--- a/internal/app/connectconformance/test_case_library_test.go
+++ b/internal/app/connectconformance/test_case_library_test.go
@@ -32,16 +32,12 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
-func asAny(t *testing.T, msg proto.Message) *anypb.Any {
-	asAny, err := anypb.New(msg)
-	require.NoError(t, err)
-	return asAny
-}
-
 func asAnySlice(t *testing.T, msgs ...proto.Message) []*anypb.Any {
 	arr := []*anypb.Any{}
 	for _, msg := range msgs {
-		arr = append(arr, asAny(t, msg))
+		asAny, err := anypb.New(msg)
+		require.NoError(t, err)
+		arr = append(arr, asAny)
 	}
 	return arr
 }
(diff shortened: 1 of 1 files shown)


smaye81:
@jhump I tried to refactor these with JSON like we discussed but that made the test pretty large since these are pretty complex objects. I tried a different approach to see if I could make them a bit more self-contained. Let me know what you think.

jhump (approved):
It does read a little better, so feel free to merge.

But I was a bit surprised that this change is net positive for new LOC, so it's larger than the previous formulation 🤷

smaye81:
It is yeah 😞  because there's no reuse of code any more and things are duplicated. If this was in the JSON format though, it would have been much larger which is why that approach scared me away. Not sure if there's an easier way. These are pretty complex objects and there's a lot of scenarios to cover 🤷‍♂️

jhump:
> If this was in the JSON format though, it would have been much larger which is why that approach scared me away. 

I don't quite follow this. The JSON format is strictly _less_ verbose than using Go struct literals for everything. The JSON notation for an object is less verbose (because you don't need to precede every value with its type), and you can omit error handling for "any" messages. Maybe you're envisioning something different than I am when imagining the use of JSON here?

Anyhow, what's here is good. I don't think it's worth a lot of head ache to make it more succinct, so don't take the above comment as a dare to use JSON 😆. I could also be just plain wrong and what I'm imagining not do-able or not better.

Ship it!

PR #751, opened 2024-01-05
Add gRPC-web client to tests

smaye81:
This adds the [gRPC-web client](https://github.com/grpc/grpc-web) implementation to the conformance tests. It runs the client against both the Connect reference server as well as the gRPC-Go server to verify consistency across the implementations.


buf.gen.yaml
```diff
@@ -13,3 +13,12 @@ plugins:
   - plugin: buf.build/grpc/go:v1.3.0
     out: internal/gen/proto/go
     opt: paths=source_relative
+  - plugin: buf.build/protocolbuffers/js:v3.21.2
+    out: grpcwebclient/gen/proto
+    opt: import_style=commonjs,binary
+  - plugin: buf.build/bufbuild/es
+    out: grpcwebclient/gen/proto/es
+    opt: target=ts
```
smaye81, line 21: The gRPC-web client expects code generated by the `protocolbuffers/js` plugin, but the code generated by Protobuf-ES is much easier and more intuitive to work with. So, what I'm actually doing is generating both and only using the `protocolbuffers/js` code in the actual client implementation when it needs to invoke calls and get responses. If there's a better way to do this, I'm down to change it.

grpcwebclient/impl.ts
```diff
@@ -0,0 +1,296 @@
+// Copyright 2023 The Connect Authors
```
smaye81, line 1: This is the actual client implementation (the equivalent of `impl.go` for the other Go clients). All the other ancillary TypeScript files are to communicate with the runner.

grpcwebclient/impl.ts
```diff
@@ -0,0 +1,296 @@
...
+  req: ClientCompatRequest,
+): Promise<ClientResponseResult> {
+  const msg = req.requestMessages[0];
+  const uReq = new UnaryRequest();
+  if (!msg.unpackTo(uReq)) {
+    throw new Error("Could not unpack request message to unary request");
+  }
+
+  // Convert from Protobuf-ES into the gRPC-web compatible library
+  const ur = UnaryRequestGoog.deserializeBinary(uReq.toBinary());
```
smaye81, line 102: re: above comment about using both plugins -- this is where I convert the Protobuf-ES requests to the `protocolbuffers/js` requests. I also hate the `Goog` suffix, so if there's a better way to name them, lmk.

internal/app/connectconformance/testsuites/grpc-web-client.yaml
```diff
@@ -0,0 +1,570 @@
+name: gRPC-web
```
smaye81, line 1: I created this file because running the client through the regular test files causes a few issues. The common thread for all of these is that there are some failures that are specific to the gRPC-web client, but _not_ with Connect via gRPC-web and we have no way to exclude the one but not the other.

* The biggest obstacle is cancellations with server-streaming. We can't really test cancellations because of the issues around `mode=grpcweb` and `mode=grpcwebtext` and that Connect doesn't support the latter (see docs [here](https://github.com/grpc/grpc-web?tab=readme-ov-file#streaming-support) and [here](https://connectrpc.com/docs/go/grpc-compatibility/#handlers)). Typically, we could just remove gRPC-web from the `relevantProtocols` list in the test file, but, we _can_ test cancellations with the Connect client using the gRPC-web protocol so that's a viable solution. We would need some way to say 'run these tests over gRPC-web using the Connect client only'.
* There is an issue with timeouts also (see [TODO](https://github.com/connectrpc/conformance/pull/751/files#diff-a8509409d477ca9786a1bcee8d03bed9c0e9e47501c85f657705510de01ff0a0R264) in this file). For some reason, timeouts of 200ms are flaky unless the server response delay is 1s or more. I'm not entirely sure if this is something on the gRPC-web side or how Connect is calculating the `Grpc-Timeout` header, but the same reason applies as cancellations -- we can test timeouts with a lower response delay using the Connect client via gRPC-web. We could increate the response delay in `timeouts.yaml` but we should probably figure out the actual issue first and just keep them isolated here (at least IMO).
* The `client_message_size.yaml` tests are hanging while trying to read the expanded requests from stdin (see TODO). For some reason, the `readReqBuffers` function in `conformance-runner.ts` is not reading these values properly. The consequence is the same as the above. This seems to only affect the gRPC-web client so we can't exclude the gRPC-web protocol from that file.
* Finally, we have the issue with trailers and trailer values with the same name ([TODO](https://github.com/connectrpc/conformance/pull/751/files#diff-a8509409d477ca9786a1bcee8d03bed9c0e9e47501c85f657705510de01ff0a0R28)). As discussed offline, we could mitigate that by simply creating a `duplicate_metadata.yaml`, but even then, we have the same problem of that working with Connect via gRPC-web vs. not working with the actual gRPC-web client.

internal/app/connectconformance/testsuites/grpc-web-client.yaml
```diff
@@ -0,0 +1,576 @@
...
+relevantHttpVersions:
+  - HTTP_VERSION_1
+relevantCodecs:
+  - CODEC_PROTO
+relevantProtocols:
+  - PROTOCOL_GRPC_WEB
+relevantCompressions:
+  - COMPRESSION_IDENTITY
```
smaye81, line 9: Should we only run with identity?

Makefile
```diff
@@ -103,6 +104,11 @@ runservertests: $(BIN)/connectconformance $(BIN)/referenceserver $(BIN)/grpcserv
 runclienttests: $(BIN)/connectconformance $(BIN)/referenceclient $(BIN)/grpcclient
 	$(BIN)/connectconformance -v --conf ./testdata/reference-impls-config.yaml --mode client -- $(BIN)/referenceclient
 	$(BIN)/connectconformance -v --conf ./testdata/grpc-impls-config.yaml --mode client -- $(BIN)/grpcclient
+	$(MAKE) rungrpcweb
+
+.PHONY: rungrpcweb
+rungrpcweb: generate $(BIN)/connectconformance
+	cd grpcwebclient && npm run conformance
```
smaye81, line 111: Not sure of the best place to put the code. Didn't want it mixed in with the Go code, so lmk if you think there's a better place.

internal/app/connectconformance/testsuites/grpc-web-client.yaml
smaye81: One way we could fix the issue of 'this works with Connect over gRPC-web but not with the gRPC-web client' is to use an expected failures file that contains the test we expect to fail with the gRPC-web client only.

.gitignore
```diff
@@ -3,3 +3,4 @@
 *.svg
 cover.out
 dist/
+node_modules
```
jhump, line 6: This is a directory, no? How about a trailing slash, just for consistency with the other directory names in this list.

```suggestion
node_modules/
```

buf.gen.yaml
jhump: So the client code you have is mixing both protobuf-es and protocolbuffers-js? That seems a little odd... Is it easy to convert between then in the client code I guess? It might be nicer to just use protocolbuffers-js, just so the setup and the code is a little simpler. But if that API and generated code is really that unpleasant, it's probably worth at least adding some comments -- maybe here in this YAML file but maybe also in a README.md for the grpcwebclient folder that describes its purpose and why the looks the way it does.

grpcwebclient/conformance-runner.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+        runnerCloseResolve();
+      });
+    });
+    await runnerClose.finally(() => server.close());
+
+    expect(true).toBe(true);
+  });
+});
+async function run(socket: net.Socket, invokeScript: string) {
```
jhump, line 108: nit: mind adding a blank line above this, so the top-level elements are little more discernible at a glance?

internal/app/connectconformance/testsuites/grpc-web-client.yaml
jhump: Seems like you don't really need to be so specific with these relevant aspects of these cases since they are redundant with what's in the supported features in `testdata/grpc-web-client-impl-config.yaml`.

testdata/grpc-web-client-impl-config.yaml
```diff
@@ -0,0 +1,13 @@
+features:
+  versions:
+    - HTTP_VERSION_1
+  protocols:
+    - PROTOCOL_GRPC_WEB
+  codecs:
+    - CODEC_PROTO
+  compressions:
+    - COMPRESSION_IDENTITY
```
jhump, line 9: From above thread: this should include gzip if the grpcweb client can handle compression. Maybe a TODO for now (you can put TODOs in this file -- it's yaml, so comments are allowed).

buf.gen.yaml
smaye81: Unfortunately that's correct. In addition to the unpleasantness of the API there is a bit of a showstopper in that there doesn't appear to be any way to convert an object _from_ JSON. There is only a `toObject` function and we need the ability to convert from JSON because the Webdriver API returns us a JSON string. There's probably ways to work around this, but in the end, it just felt easier to use Protobuf-ES everywhere and only convert to/from protocolbuffers/js when interacting with the gRPC-web client API.

I documented this in a comment in both `buf.gen.yaml` and a new README for the `grpcwebclient` directory. Hopefully this is sufficient.

@timostamm if you are aware of anything I'm missing here, let me know.

internal/app/connectconformance/testsuites/grpc-web-client.yaml
smaye81: Will do. I added a TODO to this `grpc-web-client.yaml` file to fix this, which I'll do in a follow-up PR.

internal/app/connectconformance/testsuites/grpc-web-client.yaml
smaye81: It seems that it does support compression by setting the `Content-Encoding` header and letting the browser do the compression. I updated the tests and the client implementation to correctly set this header under the various compression scenarios.

In addition, though, I made a change to the referenceserver checks for the `Content-Encoding` header. According to [this doc](https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md#compression), gRPC-web seems to use `Content-Encoding` and ignores `Grpc-Encoding`.

grpcwebclient/conformance-runner.ts
```diff
@@ -0,0 +1,205 @@
...
+  UnaryRequest,
+  ServerStreamRequest,
+  ClientStreamRequest,
+  BidiStreamRequest,
+  ConformancePayload_RequestInfo,
+  UnimplementedRequest,
+  ClientCompatResponse,
+);
+
+describe("Connect Conformance", () => {
```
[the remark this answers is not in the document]
smaye81: Sort of, yeah. Basically what this is all doing is launching Webdriver and then using it's unit test API to execute the conformance tests. The unit test logic you see here is what's run inside the Chrome headless shell, so it's easiest to just run it within a single unit test. It's a bit strange to see, but this single unit test is what's actually executing the conformance tests.

grpcwebclient/entry.ts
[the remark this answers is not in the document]
smaye81: This is essentially the entry point into the Webdriver browser instance. What's happening is the `conformance-runner.ts` file builds all the browser-specific code using esbuild and `entry.ts` is the build's entry point. All the browser code that stems from `entry.ts` is built into a bundle and then run inside Webdriver.

I could rename this if you think there's a better name for it?

internal/app/referenceserver/checks.go
```diff
@@ -212,13 +212,14 @@ func checkCompression(expected v1.Compression, req *http.Request, feedback *feed
...
-		case contentType == grpcContentType || contentType == grpcWebContentType ||
-			strings.HasPrefix(contentType, grpcContentTypePrefix) ||
-			strings.HasPrefix(contentType, grpcWebContentTypePrefix):
+		case contentType == grpcContentType ||
+			strings.HasPrefix(contentType, grpcContentTypePrefix):
 			encodingHeader = "grpc-encoding"
 		case strings.HasPrefix(contentType, connectStreamContentTypePrefix):
 			encodingHeader = "connect-content-encoding"
-		case strings.HasPrefix(contentType, connectUnaryContentTypePrefix):
+		case contentType == grpcWebContentType ||
```
smaye81, line 220: gRPC-web uses `Content-Encoding` and ignores `Grpc-Encoding`.  See [docs](https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md#compression).

internal/app/connectconformance/testsuites/grpc-web-client.yaml
jhump: > It does support compression, but I think we have an inconsistency with Connect via gRPC-web vs. the gRPC-web client.

Yeah, just skip the compression. Connect is doing things the right way. I guess the grpc-web Envoy plugin supports full-body compression, but that is not actually part of any of the specs, so fine to just skip compression entirely for this client.

grpcwebclient/conformance-runner.ts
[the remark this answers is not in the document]
jhump: Ah, okay. Maybe a short comment above explaining that to a reader?

grpcwebclient/entry.ts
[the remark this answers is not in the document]
jhump: Maybe a comment would help readers understand it.

grpcwebclient/impl.ts
```diff
@@ -71,6 +66,30 @@ function buildMetadata(req: ClientCompatRequest): Metadata {
...
+      break;
+    case Compression.SNAPPY:
+      metadata["content-encoding"] = "snappy";
+      break;
+    case Compression.IDENTITY:
+      metadata["content-encoding"] = "identity";
+      break;
+    default:
+    // Do nothing
+  }
```
jhump, line 91: Let's back this out. This is full-body compression, which is not the expected way to do compression with the gRPC / gRPC-Web protocol.

jhump (approved):
LGTM. It would be good to get sign-off from Timo before merging since he's a much better reviewer for JS/TS style and for the browser architectural bits (if he has time to do so).

buf.gen.yaml
timostamm: It is correct that github.com/protocolbuffers/protobuf-javascript does not implement the JSON format. 

I think there are several complications going on that make it a bit difficult to follow what's going on:
1. The jasmine test framework is used, but we're not actually using it to test anything.
2. We're generating code for two different protobuf implementations, so it's a bit muddy what's under test.
3. There's an inversion of control with the jasmine test invoking the `connectconformance` command. 
4. There's IPC between the jasmine test, the `pipe` command, and also IPC with the browser.

I would try to remove the inversion of control (is it buying us anything here?). 

To transmit a test case to the browser and results back, you do not need protobuf JSON. Binary data is just an array of bytes, which is just an array of numbers, which is valid JSON.

timostamm:
> It would be good to get sign-off from Timo before merging since he's a much better reviewer for JS/TS style and for the browser architectural bits (if he has time to do so).

Commented in https://github.com/connectrpc/conformance/pull/751#discussion_r1450482131. TBH, it is quite difficult to follow, and I wasn't able to run it - `npm run conformance` raised the error ` code: 'ENOENT', syscall: 'spawn ../.tmp/bin/connectconformance'`. 

Do we have the ability to look at the tests in a local browser with the developer tools? Test failures will be very difficult to debug without.

buf.gen.yaml
smaye81: Ideally, we should be passing `conformance-runner.ts` file to the conformance binary and that .ts file would launch Webdriver, the browser, etc. I actually started down this path when first trying to write this but ran into one problem after another with launching the browser and communicating via `stdin`. In the end, I settled on the approach in https://github.com/connectrpc/connect-es/pull/952 for two reasons: 1. it works and 2. I felt like I was spending too much time tinkering around with everything for something that is arguably not entirely required to test Connect implementations.

I can revisit all of this if we'd rather it work a different way. Just not sure how much time we want to invest in that. I guess at the least I can try to remove the usage of two different Protobuf implementations. 

Let me know what you guys think though.

smaye81:
> > It would be good to get sign-off from Timo before merging since he's a much better reviewer for JS/TS style and for the browser architectural bits (if he has time to do so).
> 
> Commented in [#751 (comment)](https://github.com/connectrpc/conformance/pull/751#discussion_r1450482131). TBH, it is quite difficult to follow, and I wasn't able to run it - `npm run conformance` raised the error ` code: 'ENOENT', syscall: 'spawn ../.tmp/bin/connectconformance'`.
> 
> Do we have the ability to look at the tests in a local browser with the developer tools? Test failures will be very difficult to debug without.

Looks like you never built the conformance binary. If you run `make rungrpcweb` from the project root, it should hopefully work fine.

Also re: viewing the tests in a browser, there's nothing I'm aware of, no. I tried using Webdriver's loggers to capture this, but was not successful.

buf.gen.yaml
smaye81: Update: was able to remove Protobuf-ES and strictly use protocolbuffers/js. I'm passing the binary data to and from Webdriver now 👍

grpcwebclient/runner.ts
```diff
@@ -0,0 +1,150 @@
+// Copyright 2023-2024 The Connect Authors
```
smaye81, line 1: @timostamm I know you mentioned you didn't like the name `runner.ts`. I'm open to call this whatever.

grpcwebclient/runner.ts
```diff
@@ -0,0 +1,150 @@
...
+  ClientResponseResult,
+  ClientErrorResult,
+} from "./gen/proto/connectrpc/conformance/v1/client_compat_pb.js";
+
+export async function run() {
+  // Launch a browser. For a non-headless browser, pass false
+  const browser = await puppeteer.launch({ headless: "new" });
+  const page = await browser.newPage();
+
+  page.on("pageerror", (err) => {
```
smaye81, line 30: @timostamm I tried a few things to get this to trip, but I couldn't do it. I'm not sure if this is still needed since we have the try/catch below now?

smaye81:
@timostamm @jhump Refactored this to use Puppeteer (based on Timo and my pairing session yesterday) and it's much easier to follow. Lmk what you think.

grpcwebclient/runner.ts
timostamm: It's possible that it only triggers for uncaught errors, for example a syntax error in a script. I wouldn't worry about it too much with well-placed try-catch.

grpcwebclient/runner.ts
timostamm: You have a `bin/grpcwebclient`, which simply delegates to this file, so I'd just call it `grpcwebclient.ts`.

grpcwebclient/runner.ts
```diff
@@ -0,0 +1,150 @@
...
+import * as esbuild from "esbuild";
+import {
+  ClientCompatRequest,
+  ClientCompatResponse,
+  ClientResponseResult,
+  ClientErrorResult,
+} from "./gen/proto/connectrpc/conformance/v1/client_compat_pb.js";
+
+export async function run() {
+  // Launch a browser. For a non-headless browser, pass false
```
timostamm, line 26: 🎉

timostamm (approved):
This removes all of the complications from https://github.com/connectrpc/conformance/pull/751#discussion_r1450482131 👍 

> Looks like you never built the conformance binary.

I guess I somehow managed to do that.

PR #754, opened 2024-01-12
Modify processing of known failing tests

smaye81:
Currently, when specifying a `known-failing` file, the tests that are listed in the file are matched against the list of test cases. If there are tests or patterns that do not match, an error is thrown.

However, this causes an issue when trying to mark one of the gRPC tests as known-failing because the gRPC test names are modified via the `filterGRPCImplTestCases` function. Tests have `(grpc server impl)`, `(grpc client impl)` or `(grpc impl)` added as the penultimate value before the test name, depending on how they are run. But since this is done _after_ the known-failing tests are matched, it is impossible to mark one of these gRPC tests as failing.

So, this PR modifies the calculation of the test names a bit by creating a list of all known test permutations and then using that to match against any known-failing tests listed.

Note that there is an example `known-failing` file and a test deliberately set up to fail just to see how this works in action. These two files will be removed from this PR before it is merged.

To run with these changes, use the command:

```bash
go run cmd/connectconformance/main.go -v --known-failing internal/app/connectconformance/testsuites/known_failing.txt --mode client -- .tmp/bin/referenceclient
```

### DIFF
--- a/internal/app/connectconformance/connectconformance.go
+++ b/internal/app/connectconformance/connectconformance.go
@@ -443,52 +443,23 @@ func calcAllTestCases(testCases map[string]*conformancev1.TestCase, clientIsGRPC
 	for _, testCase := range testCases {
 		testCaseSlice = append(testCaseSlice, testCase)
 	}
+	var gRPCClientTests, gRPCServerTests, gRPCClientServerTests []*conformancev1.TestCase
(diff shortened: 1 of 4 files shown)


jhump (approved):
I think a lot of this should really be the responsibility of the test case library, but this seems good for now.

PR #757, opened 2024-01-16
Add command-line flags for more control over test cases to run and how they're handled

jhump:
This adds a few new features:
* The existing `--test-file` flag can now be specified multiple times, so one can run multiple suites at the same time, directly from a YAML file instead of from the embedded test case data. (Related TODO: it would be nice to also support a `--test-folder` flag, to point to a directory and pull in all YAML files therein.)
* The existing `--known-failing` flag is changing to be a pattern, and can be specified multiple times. To refer to a file of patterns (old behavior), prefix the value with "@".
* The `--known-flaky` flag, which indicates patterns of test case names that are known to be flaky. The difference between "known failing" and "known flaky" is that if a known flaky test happens to succeed, the conformance tests will still succeed. With a known failing test, however, if it passes, that causes the run to fail. The "known failing" set should generally be preferred so that if/when a test case is fixed, the conformance tests will enforce that you also update your list of known failing test cases.
* The `--run` flag, which indicates patterns of test case names to include in the run. If not specified, _all_ test cases that apply (per config) are included. This allows running just a single test suite, test case, or test case permutation, depending on what the user is trying to debug.
* The `--skip` flag, which is the reverse above. This indicates patterns of test case names to _not_ bother running. In general, users should prefer `--known-failing` or `--known-flaky` flags to opt-out of test cases. But this is a convenient way to simply _not run_  particular test cases when running the conformance tests.

This also makes some other minor changes and improvements in the code:
1. Moves the computation of gRPC-implementation test case permutations into the `testCaseLibrary`.
2. Renames the "known failing" files and data structures to `test_trie.go` / `testTrie`, since it's really a general-purpose trie implementation, that is now also used to implement other pattern matching (for new "known flaky", "run" and "skip" options).
3. Improved the look and consistency of error output on failure. Now it always puts _all_ error messages below the main "FAILED" line, one error per line. Previously, it would put the first error on the same line, or even print just a single line if there was only one error. These error messages are also indented to aid visually scanning the output.


cmd/connectconformance/main.go
```diff
@@ -33,6 +33,9 @@ const (
 	configFlagName        = "conf"
 	testFileFlagName      = "test-file"
 	knownFailingFlagName  = "known-failing"
+	knownFlakyFlagName    = "known-flaky"
+	runFlagName           = "run"
+	noRunFlagName         = "no-run"
```
smaye81, line 38: I feel like `skip` would be slightly clearer, but I'm not hellbent on it 🤷‍♂️

smaye81 (approved):

cmd/connectconformance/main.go
jhump: I debated that. I was looking for something that was obviously an antonym for `--run`. I considered `--include` and `--exclude`, but was worried those words weren't as clear/obvious when it comes to what these flags were for. But I can see "skip" also being obvious, even if the word by itself isn't related to the word "run".

Another thing that concerned me a little bit was how this now differs from the `--known-failing` flag, which names a file instead of being an actual pattern that can be specified multiple times. Maybe I should change them all to be consistent so that all of `--known-failing`, `--known-flaky`, `--run`, and `--skip` take a _pattern_ as their argument and can be specified multiple times. But you can prefix the argument with `@` to indicate the name of a file to read. So existing usages would have to add `@` to their `--known-failing` arguments, but then everything is more consistent.

@smaye81, WDYT?

smaye81: Sounds good to me. We'd only have one place (grpc-web) to fix the usage of `--known-failing` so that should be easy.

jhump: > We'd only have one place (grpc-web) to fix the usage of --known-failing so that should be easy.

It's actually used in connect-kotlin and connect-swift, too. But still easy to update.

PR #762, opened 2024-01-23
Test HTTP code to Connect code mappings

smaye81:
This adds some additional tests to test that the mapping of HTTP code to Connect code is correct for clients according to the protocol: https://connectrpc.com/docs/protocol#http-to-error-code

This also adds usage of `connect.ErrorWriter` to the raw response middleware (although is admittedly maybe no longer needed since the header issue has been fixed).

### DIFF
--- a/internal/app/connectconformance/test_case_library.go
+++ b/internal/app/connectconformance/test_case_library.go
@@ -545,9 +545,11 @@ func populateExpectedUnaryResponse(testCase *conformancev1.TestCase) error {
 	}
 
 	if def.RawResponse != nil {
-		fmt.Println("harr")
-		expected.Error = &conformancev1.Error{
-			Code: 2,
+		// TODO - Handle other raw response values
+		if def.RawResponse.StatusCode != 0 {
+			expected.Error = &conformancev1.Error{
+				Code: mapHTTPtoConnectCode(def.RawResponse.StatusCode),
+			}
 		}
 	} else {
 		switch respType := def.Response.(type) {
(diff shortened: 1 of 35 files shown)


internal/app/referenceserver/raw_response.go
```diff
@@ -39,12 +39,22 @@ type rawResponseKey struct{}
...
+			msg := "missing x-test-case-name header"
+			if errorWriter.IsSupported(req) {
+				invalidArg := connect.NewError(connect.CodeInvalidArgument, errors.New(msg))
+				err := errorWriter.Write(respWriter, req, invalidArg)
+				if err != nil {
+					http.Error(respWriter, err.Error(), http.StatusInternalServerError)
+				}
+			} else {
+				http.Error(respWriter, msg, http.StatusBadRequest)
+			}
```
jhump, line 57: This part LGTM. If you want to pull this out into a separate PR, we can go ahead and get it merged. (Though not a big deal if you just want to leave it here -- if this isn't obviously valuable outside of this PR, then maybe no need to do the work to carve out separate PR for it...).

internal/app/connectconformance/test_case_library.go
```diff
@@ -566,31 +566,43 @@ func populateExpectedUnaryResponse(testCase *conformancev1.TestCase) error {
...
-			RequestInfo: reqInfo,
+	if def.RawResponse != nil {
+		// TODO - Handle other raw response values
+		// If an HTTP error status was specified in the raw response, then the test case is
+		// forcing the return of an HTTP error, which should be mapped to a Connect error
+		// according to the protocol.
+		if def.RawResponse.StatusCode >= 400 && def.RawResponse.StatusCode <= 599 {
+			expected.Error = &conformancev1.Error{
+				Code: mapHTTPtoConnectCode(def.RawResponse.StatusCode),
+			}
```
smaye81, line 577: Not quite sure about this whole block. I know we discussed that you could specify a raw response _and_ the standard response definition fields, but I was curious about two things:

- The [proto comment](https://github.com/connectrpc/conformance/blob/main/proto/connectrpc/conformance/v1/service.proto#L168) mentions it should be one or the other in test cases
- If the comment is incorrect, then how should we best handle the case where an HTTP error code is specified in the raw response along with a defined `Error` to be returned? Should the defined `Error` override the raw response error in `expected`?

jhump: I don't have a strong opinion here. I can see, for clarity, it would be best to just disallow the standard response definition.

But I also think, for consistency, there are things we'll want to tweak in how raw _requests_ work (since it is currently _necessary_ to specify both there).

Having said all that, I don't think we want this block. For raw responses, we should just require that the expected response be specified in the YAML file. That makes the test cases much more clear.

internal/app/connectconformance/testsuites/http_to_connect_code.yaml
```diff
@@ -0,0 +1,158 @@
...
+    testName: bad request
+    service: connectrpc.conformance.v1.ConformanceService
+    method: Unary
+    streamType: STREAM_TYPE_UNARY
+    requestMessages:
+    - "@type": type.googleapis.com/connectrpc.conformance.v1.UnaryRequest
+      responseDefinition:
+        responseData: "dGVzdCByZXNwb25zZQ=="
+        rawResponse:
+          statusCode: 400
```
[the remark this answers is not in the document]
smaye81: Yeah that's fair. I agree. Updated the test cases here to use the explicit expected response and added a validation to make sure if a raw response is specified that an explicit response is also specified.

jhump (approved):
LGTM!

PR #768, opened 2024-01-30
Add wire tracer capabilities to reference client

smaye81:
This piggybacks off of the tracer framework to add wire tracing to the reference client. Broadly, this implements the `Collector` interface and intercepts the `Complete` function invocation to sniff the built trace and acquire wire details.

The details tracked are:
- Actual HTTP status code
- Actual trailers on the wire
- Connect Raw Error Details (if applicable)

Note that the tests to verify raw error details trailers are not yet in place to keep the complexity of this PR lower. They will be added in a follow-up PR.


internal/app/connectconformance/testsuites/errors.yaml
```diff
@@ -412,6 +412,31 @@ testCases:
         responseTrailers:
         - name: x-custom-trailer
           value: ["bing"]
+# Client Stream Tests -----------------------------------------------------------
```
smaye81, line 415: Admittedly a drive-by change. There were no client stream error tests in this file.

internal/app/referenceclient/client.go
```diff
@@ -216,9 +216,11 @@ func invoke(ctx context.Context, req *v1.ClientCompatRequest, trace *tracer.Trac
...
 		return nil, errors.New("an HTTP version must be specified")
 	}
-	if trace != nil {
-		transport = tracer.TracingRoundTripper(transport, trace)
-	}
+
+	// Create a new TracingRoundTripper with our WireTracer so that the tests can trace values on the
+	// wire. Note that 'trace' could be nil, in which case, any error traces will
+	// simply not be printed.
+	transport = tracer.TracingRoundTripper(transport, NewWireTracer(trace))
```
smaye81, line 223: Note that this now creates the `TracingRoundTripper` every time since it's needed for the wire details. If `tracer` is nil, it just bypasses the trace printing.

internal/tracer/middleware.go
```diff
@@ -39,12 +40,24 @@ func TracingRoundTripper(transport http.RoundTripper, collector Collector) http.
...
+		}
+		resp.Body = io.NopCloser(bytes.NewReader(body))
+		builder.add(&ResponseStart{
+			Response: resp,
+		})
+
 		respClone := *resp
-		respClone.Body = newReader(resp.Header, resp.Body, false, builder, cancel)
+		respClone.Body = newReader(resp.Header, io.NopCloser(bytes.NewReader(body)), false, builder, cancel)
+
```
smaye81, line 60: I don't love this, but not sure if there's a better way. I considered two other approaches:

* Storing the `body` data on the `ResponseStart` event so that it could be accessed by the `Complete` function override. The response body is needed there for reading unary details. 
* Creating a reusable reader that could be read 1-to-many times, but that kind of felt like overkill. 

We basically need the body data twice -- once inside the `Complete` function and once inside the `trace.Reader`. So, I opted for reading it once from the network here and then using an in-memory reader in both places. Def open to a better way though.

smaye81: Oh, also, the `io.ReadAll` call seems to hang for full-duplex streams that return a result. It works fine for every other type (including half-duplex AND full-duplex that do not return a response). Not sure if you're aware of anything that would be causing that.

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+// Lock acquires the internal lock for setting the wire details. This should
+// always be called before calling Set.
+func (w *WireWrapper) Lock() {
+	w.mtx.Lock()
+}
+
+// Lock releases the internal lock for setting the wire details.
+func (w *WireWrapper) Unlock() {
+	w.mtx.Unlock()
+}
```
jhump, line 68: It is almost never appropriate to expose internal mutexes this way. Instead, expose the higher-level operation as a method, and have the lock/unlock be internal implementation details.

In this case, it isn't necessary at all. The one place it's used can omit the locks, and instead you would put the synchornization into the `Set` method. There's no reason for it to hold the lock while computing the wire details from the trace.

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// NewWireTracer returns a new wire tracer with the given tracer.
+// If trace is nil, all tracer operations will be bypassed.
+func NewWireTracer(trace *tracer.Tracer) *WireTracer {
+	return &WireTracer{
+		tracer: trace,
+	}
+}
```
jhump, line 159: For these things that are only used from one place and are just implementation details of `main`, I'd forego the ceremony of exported types, exported factory functions, etc. So all of these types and functions can be unexported/lowercase, and trivial stuff that is used exactly once, like this function, could be inlined.

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Lock releases the internal lock for setting the wire details.
+func (w *WireWrapper) Unlock() {
+	w.mtx.Unlock()
+}
+
+// Set sets the wire details. Note that calls to Set should always be
+// preceded by calls to Lock.
+func (w *WireWrapper) Set(details *WireDetails) {
+	w.val = details
```
jhump, line 73: Oof, this seems bad to provide an exported function that provides unsynchronized access to this value. This should lock before and unlock after.

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Get returns the wire details.
+func (w *WireWrapper) Get() *WireDetails {
+	w.mtx.Lock()
+	defer w.mtx.Unlock()
+	if w.val == nil {
+		return nil
+	}
+	return w.val
```
jhump, line 56: No need for a separate nil check.
```suggestion
	return w.val
```

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+	// The actual trailers observed.
+	Trailers []*v1.Header
+	// The actual JSON observed on the wire in case of an error from a Connect server.
+	// This will only be non-nil if the protocol is Connect and an error occurred.
+	ConnectErrorRaw *structpb.Struct
+}
+
+type WireWrapper struct {
+	val *WireDetails
+	mtx sync.Mutex
```
jhump, line 46: Typical convention is to put the mutex before the fields that it guards.

However, for this, I'd probably not use a mutex at all but instead use an `atomic.Pointer[WireDetails]`. That will be more efficient since the only thing you need to do under lock is read one value or write one value.

internal/app/referenceclient/client.go
jhump: No better way; this is exactly what I would expect to see.

internal/app/referenceclient/impl.go
```diff
@@ -117,8 +118,21 @@ func (i *invoker) unary(
 	var trailers []*v1.Header
 	payloads := make([]*v1.ConformancePayload, 0, 1)
 
+	ctx, wire := WithWireCapture(ctx)
```
jhump, line 121: I think a more intuitive abstraction for this would be something like so:
```go
ctx := withWireCapture(ctx)

// perform RPC using ctx

// then at the very end
wireDetails := getWireDetails(ctx)
// ... do stuff with wire details
```
That way the wire wrapper never needs to be exposed and doesn't need any methods -- it's just an implementation detail of your wire tracer.

Even if you keep the shape the way it is, I think this means you don't need to provide a `Set` method -- only `Get` needs to be exposed.

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+	"sync"
+
+	"connectrpc.com/conformance/internal"
+	v1 "connectrpc.com/conformance/internal/gen/proto/go/connectrpc/conformance/v1"
+	"connectrpc.com/conformance/internal/tracer"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+type wireKey struct{}
```
jhump, line 31: maybe name it `wireCtxKey` to make it more clear that this is used as a key for a context value? (and/or add a doc comment?)

internal/app/referenceclient/wire_tracer.go
```diff
@@ -0,0 +1,159 @@
...
+
+// WireDetails encapsulates the wire details to track for a roundtrip.
+type WireDetails struct {
+	// The actual HTTP status code observed.
+	StatusCode int32
+	// The actual trailers observed.
+	Trailers []*v1.Header
+	// The actual JSON observed on the wire in case of an error from a Connect server.
+	// This will only be non-nil if the protocol is Connect and an error occurred.
+	ConnectErrorRaw *structpb.Struct
```
jhump, line 41: Aside: I got another one we'll want to capture. Not that you need to add it here; I can come back and add it soon.

But we will also want to capture the raw trailers in a gRPC-Web end-stream message. Because we'll want to validate that they were all lower-case on the wire.

internal/tracer/middleware.go
```diff
@@ -39,12 +40,24 @@ func TracingRoundTripper(transport http.RoundTripper, collector Collector) http.
...
 		}
-		builder.add(&ResponseStart{Response: resp})
+
+		defer resp.Body.Close()
+		// Read the response body in full and then replace it with a new in-memory reader
+		// so that we can read the contents multiple times in any trace or middleware.
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
```
jhump, line 52: Doesn't this break full-duplex bidi streams? I don't think you can do this.

What I was thinking is that instead, over in the reference client only (_not_ in the tracer), you'd have middleware that optionally captures the body. And it shouldn't do so by trying to read the entire response -- it needs to replace `Body` with a new implementation that will capture the bytes as they are read, like into a `*bytes.Buffer`. In fact, you could even store the `*bytes.Buffer` in the wire wrapper struct, and the middleware only needs to activate (to replace the body) when there is a wire wrapper in context.

internal/tracer/middleware.go
jhump: Yep, you can't do this in a round-tripper. See the comment I just added above. You need middleware to replace the `Response.Body` with an `io.ReadCloser` that will capture the data as it is read.

internal/app/connectconformance/results.go
```diff
@@ -199,6 +199,14 @@ func (r *testResults) assert(testCase string, expected, actual *conformancev1.Cl
 		}
 	}
 
+	// If a status code is specified in the expected result, then verify it against the actual status code
```
smaye81, line 202: Temporary for testing - added checks on the status code

internal/app/connectconformance/test_case_library.go
```diff
@@ -600,6 +600,7 @@ func populateExpectedUnaryResponse(testCase *conformancev1.TestCase) error {
 			payload.Data = respType.ResponseData
 		}
 		expected.Payloads = []*conformancev1.ConformancePayload{payload}
+		expected.ActualStatusCode = 200
```
smaye81, line 603: Temporary for testing

internal/app/referenceclient/impl.go
```diff
@@ -117,8 +119,24 @@ func (i *invoker) unary(
...
 
+	ctx = withWireCapture(ctx)
+
 	// Invoke the Unary call
 	resp, err := i.client.Unary(ctx, request)
+
+	var actualStatusCode int32
+	var connectErrorRaw *structpb.Struct
+	var actualTrailers []*v1.Header
+	fmt.Fprintf(os.Stderr, "Getting wire details in client impl for %s\n\n", req.TestName)
```
smaye81, line 130: Temp for testing - occasionally this and line 137 will print before setting the value in `wire_tracer`. It will be evident because a test will fail with status code 0.

internal/app/referenceclient/tracer.go
```diff
@@ -104,7 +94,7 @@ func (t *WireTracer) Complete(trace tracer.Trace) {
 				if statusCode != 200 {
 					// If this is a unary request, then use the entire response body
 					// as the wire error details.
-					body, err := io.ReadAll(trace.Response.Body)
+					body, err := io.ReadAll(wrapper.buf)
```
jhump, line 97: This is not really safe -- the `Complete` call could be from a different goroutine from the one that is writing to `buf`, so you must synchronize. And just making `buf` an atomic pointer isn't sufficient because it's not just the pointer to `bytes.Buffer` but also the contents of the buffer.

I think it would be safe to instead have this `Complete` call merely store the trace in the `wireWrapper` atomic pointer, and then delay computing the wire details -- from the trace _and_ the buffer -- until the call to `getWireDetails`.

I _think_ that would be thread-safe, since the call to `getWireDetails` should always happen at the end of the RPC, after the whole thing is done and the entire response body has been read. So at that point, there should be no other concurrent activity the handler could be racing with.

internal/app/referenceclient/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+package referenceclient
+
+import (
+	"io"
+	"net/http"
+
+	"connectrpc.com/conformance/internal/tracer"
+)
+
+type wireInterceptor struct {
```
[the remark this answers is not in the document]
smaye81: I originally had it that way and then felt uneasy about that file containing so many various types and functions. I moved them back though. Let me know if it seems off.

internal/app/referenceclient/tracer.go
smaye81: Yeah this actually also allows me to return an `error`, which in turn enables propagating this error back to the surface. I like this approach much better.

internal/app/referenceclient/impl.go
```diff
@@ -341,32 +371,51 @@ func (i *invoker) clientStream(
...
+	ctx = withWireCapture(ctx)
+
 	stream := i.client.BidiStream(ctx)
 	defer func() {
+		wireDetails, e := getWireDetails(ctx)
+		if e != nil {
+			result = nil
+			err = e
+			return
+		}
```
smaye81, line 410: I want to call `getWireDetails` in the `defer` function because it's much easier to reason about in the bidi code. But, wasn't sure if this was the best way to handle the error.

jhump: A few things:
1. I think you only want to call `getWireDetails` when `err == nil`. If this function is already returning an error (and not a result), `getWireDetails` is about guaranteed to fail, and this would also overwrite the existing error that is being returned.
2. When you get an error, I think you'd want to instead store it in [`ClientCompatResponse.feedback`](https://buf.build/connectrpc/conformance/docs/main:connectrpc.conformance.v1#connectrpc.conformance.v1.ClientCompatResponse). Admittedly, that's a bit of a pain to do with the current structure. Maybe you need something like a `internal.Printer` to collect feedback messages and pass that into `invoke` (and then into this method). Then this function could write the error message into that printer. Then the caller, when it's creating the `ClientCompatResponse` (using either the result or error that this returns), can add those messages to that field. _OR_ we could just move the `feedback` field _into_ the `ClientResponseResult` message. WDYT? (I added it as a "just in case", to have feedback from the reference client like we have feedback from the reference server; but it's not yet used.)

jhump: Though... I'm slightly worried that maybe an error getting those details isn't something that should fail a conformance test case. Like maybe it indicates a bug in the reference client _not_ in the server under test. So in that case, you might just eat it and put a TODO for what to do with it.

internal/app/connectconformance/results.go
```diff
@@ -191,11 +191,17 @@ func (r *testResults) assert(testCase string, expected, actual *conformancev1.Cl
...
-		if diff != "" {
-			errs = append(errs, fmt.Errorf("raw Connect error does not match: - wanted, + got\n%s", diff))
+	expectedWire := expected.WireDetails
+	actualWire := actual.WireDetails
+	if expectedWire != nil && actualWire != nil {
+		// TODO - Add comparison (and tests) for connecterrorraw and actual http trailers
+
+		// if diff := cmp.Diff(expectedWire.ConnectErrorRaw, actualWire.ConnectErrorRaw, protocmp.Transform()); diff != "" {
+		// 	errs = append(errs, fmt.Errorf("raw Connect error does not match: - wanted, + got\n%s", diff))
+		// }
```
smaye81, line 201: Needed to comment out this check for now since we validate all details when message is not nil.

internal/app/referenceclient/impl.go
```diff
@@ -386,53 +350,35 @@ func (i *invoker) clientStream(
...
-	result = &v1.ClientResponseResult{
-		ConnectErrorRaw: nil, // TODO
-	}
+	result = &v1.ClientResponseResult{}
 
 	ctx = withWireCapture(ctx)
 
 	stream := i.client.BidiStream(ctx)
 	defer func() {
 		if result != nil {
```
jhump, line 376: I'd invert this so you can un-indent the rest of the function. So basically this at the beginning:
```go
if err != nil {
    return
}
```

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -140,20 +140,10 @@ message ClientResponseResult {
...
-  // the actual JSON received on the wire.
-  google.protobuf.Struct connect_error_raw = 7;
-  // Any HTTP trailers observed after the response body. These do NOT
-  // include trailers that conveyed via the body, as done in the gRPC-Web
-  // and Connect streaming protocols.
-  repeated Header actual_http_trailers = 8;
+  // The following field is only used by the reference client. If
+  // you are implementing a client under test, you may ignore this
+  // field and should not populate it.
+  WireDetails wire_details = 6;
```
jhump, line 146: I think you'll have to temporarily disable the `buf breaking` check and then re-enable it in the next PR.

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -162,3 +152,15 @@ message ClientResponseResult {
...
+message WireDetails {
+  // The HTTP status code of the response.
+  int32 actual_status_code = 1;
+  // When processing an error from a Connect server, this should contain
+  // the actual JSON received on the wire.
+  google.protobuf.Struct connect_error_raw = 2;
+  // Any HTTP trailers observed after the response body. These do NOT
+  // include trailers that conveyed via the body, as done in the gRPC-Web
+  // and Connect streaming protocols.
+  repeated Header actual_http_trailers = 3;
```
jhump, line 165: Consider moving the `feedback` field from `ClientCompatResponse` to here. Then you could also put any error message from trying to compute wire details into that field.

smaye81: In that case then, we should probably rename `WireDetails` to something else since it no longer is just details about the wire. Unless we consider the feedback a 'wire' detail.  But, maybe something to reflect it's a reference-client only field i.e. `ReferenceClientMetadata` or something.

jhump: I think the feedback would mostly be about wire details, so it would be fine to keep the name.

I don't actually have any examples of how we'd use and added it just for symmetry with the server. The server currently just uses it to complain if the incoming request doesn't have the expected properties (like HTTP version, protocol, codec, compression, etc). Technically, those are really wire details. Semantics are totally handled via the normal assertions in the test runner that compare the expected and actual responses.

.github/workflows/buf.yaml
```diff
@@ -16,21 +16,23 @@ jobs:
...
-        with:
-          input: proto
-          against: 'buf.build/connectrpc/conformance'
+  # TODO - Reenable after pushing breaking proto changes to BSR
+  # breaking:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v4
+  #     - uses: bufbuild/buf-setup-action@v1
+  #     - uses: bufbuild/buf-breaking-action@v1.1.3
```
jhump, line 25: I think _this_ is the only line you actually need to comment out, if you wanted to simplify this a little. Fine to leave as is, too.

internal/tracer/builder.go
```diff
@@ -41,6 +41,7 @@ func newBuilder(req *http.Request, collector Collector) *builder {
 		start:     time.Now(),
 		trace: Trace{
 			TestName: testName,
+			Request:  req,
```
jhump, line 44: Good catch, thanks for fixing this!

jhump (approved):
LGTM

PR #771, opened 2024-02-05
Relaxes assertions on query parameters

jhump:
Previously, this would add an expected query parameter for the `message` param, with an encoded value. But the problem is that not all clients would necessarily encode the message in exactly the same way, and that's fine. Some will choose to base64-encode (which is usually more efficient with binary payloads). Also, there's no _canonical_ protobuf binary format, so there could be subtle differences in the byte output, even for equivalent messages. Finally, there's no _canonical_ encoding for compressed payloads -- different default compression settings or other subtle differences in implementation can lead to more than one valid compressed encoding of the same data.

### DIFF
--- a/internal/app/connectconformance/results.go
+++ b/internal/app/connectconformance/results.go
@@ -159,6 +159,9 @@ func (r *testResults) failed(testCase string, err *conformancev1.ClientErrorResu
 func (r *testResults) assert(testCase string, expected, actual *conformancev1.ClientResponseResult) {
 	var errs multiErrors
 
+	errs = append(errs, checkError(expected.Error, actual.Error)...)
+	errs = append(errs, checkPayloads(expected.Payloads, actual.Payloads)...)
+
 	// TODO - This check is for trailers-only and should really only apply to gRPC and gRPC-Web protocols.
 	// Previously, it checked for error != nil, which is compliant with gRPC. But gRPC-Web does trailers-only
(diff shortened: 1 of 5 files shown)


internal/app/connectconformance/results.go
```diff
@@ -159,6 +159,9 @@ func (r *testResults) failed(testCase string, err *conformancev1.ClientErrorResu
 func (r *testResults) assert(testCase string, expected, actual *conformancev1.ClientResponseResult) {
 	var errs multiErrors
 
+	errs = append(errs, checkError(expected.Error, actual.Error)...)
+	errs = append(errs, checkPayloads(expected.Payloads, actual.Payloads)...)
+
```
jhump, line 164: This was a slight UX improvement -- often, the reason a test case fails is because an unexpected error occurs. Previously when that happened, the test case would complain about response headers and trailers before getting to the "meat" of the problem: the error. So this reorders the checks so it is more likely the most relevant errors are reported first.

jhump:
cc @rebello95, this should allow you to remove the "known failing" file in connect-swift. This will be in an upcoming v1.0.0-rc3 of the conformance tests (along with some new test cases).

rebello95:
Thanks @jhump!

smaye81 (approved):
We could probably remove `MarshalStable` from codec.go now since all the usages were removed here.

PR #772, opened 2024-02-06
Verify deadlines are correctly propagated

jhump:
I noticed in the code the other day that we _do_ verify the timeout, by having the server echo back the timeout it observed and then having the test runner check that against the timeout that the client should have sent.

So I was surprised that this didn't manifest as failed tests in connect-kotlin, which I now know (from getting pretty deep into the code) does not work -- it simply never tried to set deadline headers. (And it's not easy to add with it's current implementation structure, so I'll probably leave it as "known failing" until we can fix other problems with timeouts.)

Anyhow, when digging into why this didn't tickle any failure in connect-kotlin, it's because the only test cases that actually set the timeout always fail with a timeout error (so the client never gets back a response and cannot see the timeout that the server observed).

So this adds a new test suite to verify deadlines are correctly propagated. And it also required some fixes in the grpc-go server implementation in this repo, which was not reporting them correctly, which caused failures in the `make runservertests` target.

### DIFF
--- a/internal/app/connectconformance/results.go
+++ b/internal/app/connectconformance/results.go
@@ -36,7 +36,7 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
-const timeoutCheckGracePeriodMillis = 500
+const timeoutCheckGracePeriodMillis = 100
 
 // testResults represents the results of running conformance tests. It accumulates
 // the state of passed and failed test cases and also reports failures to a given
--- a/internal/app/grpcserver/impl.go
+++ b/internal/app/grpcserver/impl.go
@@ -62,11 +62,12 @@ func (c *conformanceServiceServer) Unary(
 		if err := grpc.SetTrailer(ctx, trailerMD); err != nil {
 			return nil, err
 		}
-		time.Sleep((time.Duration(req.ResponseDefinition.ResponseDelayMs) * time.Millisecond))
+		time.Sleep(time.Duration(req.ResponseDefinition.ResponseDelayMs) * time.Millisecond)
 	}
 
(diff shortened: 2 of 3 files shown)


internal/app/connectconformance/results.go
```diff
@@ -36,7 +36,7 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
-const timeoutCheckGracePeriodMillis = 500
+const timeoutCheckGracePeriodMillis = 100
```
jhump, line 39: Give the tolerances in actual timeout tests are on the order of 100 millis, the value here seemed egregiously generous. So I tightened it down to 100 milliseconds. This exists because some implementations (including connect-go) start the timeout clock when a deadline is configured, not when the RPC is actually sent.

So this grace period (admittedly still a bit generous) allows up to 100 milliseconds to elapse in the client implementation between the time the timeout is configured and the time it is written into RPC header metadata.

internal/app/grpcserver/impl.go
```diff
@@ -62,11 +62,12 @@ func (c *conformanceServiceServer) Unary(
 		if err := grpc.SetTrailer(ctx, trailerMD); err != nil {
 			return nil, err
 		}
-		time.Sleep((time.Duration(req.ResponseDefinition.ResponseDelayMs) * time.Millisecond))
+		time.Sleep(time.Duration(req.ResponseDefinition.ResponseDelayMs) * time.Millisecond)
```
jhump, line 65: This change ("remove redundant parentheses") was suggested by GoLand, so I went ahead and accepted here and throughout.

smaye81 (approved):

PR #774, opened 2024-02-07
Add tests for unexpected raw responses

smaye81:
This adds tests for unexpected values such as:

* Unmapped HTTP code
* Nonexistent HTTP code
* Invalid Content-Type
* Invalid Content-Encoding


jhump:
@smaye81, I had updates this to test it against latest connect-go (on main, not from a release). I've left the content-type test commented out, since it won't pass for real until we cut a connect-go release and then update this repo. But everything else is good. The only changes I made were to reconcile this with the recent changes to the location and naming conventions of test suite files. I also split the one new file into two since the last test case is using the Connect unary protocol in the response, so it seemed like it belongs in a Connect-specific file. (The other test cases work with all protocols.)

Anyhow, I think we should go ahead and move this out of draft and merge it. The commented out test has a TODO, and we can address it in a later PR (hopefully tomorrow afternoon).

jhump (approved):

smaye81:
Sounds good. Thanks for updating!

PR #776, opened 2024-02-07
Move feedback into ClientCompatResponse

smaye81:
This moves the `feedback` field into the `ClientCompatResponse` so that it is much easier to report feedback on the client-side. This way client implementations can simply set the `Feedback` field when returning a response, which will then be written out by the server runner.


internal/app/referenceclient/impl.go
```diff
@@ -376,8 +388,10 @@ func (i *invoker) bidiStream(
 		if err != nil {
 			return
 		}
+		wireDetails, feedback := getWireDetails(ctx)
 
-		result.WireDetails = getWireDetails(ctx)
+		result.WireDetails = wireDetails
+		result.Feedback = feedback
```
jhump, line 394: nit: this can be simplified to one line
```go
result.WireDetails, result.Feedback = getWireDetails(ctx)
```

smaye81: 🤯 didn't know that

PR #780, opened 2024-02-08
Reference client returns actual gRPC-Web trailer data, for new test runner checks

jhump:


internal/app/connectconformance/results.go
```diff
@@ -547,3 +535,118 @@ func indent(s string) string {
 	}
 	return strings.Join(lines, "\n")
 }
+
+func checkWireDetails(testCaseReq *conformancev1.ClientCompatRequest, expected, actual *conformancev1.ClientResponseResult) multiErrors {
+	expectedWire := expected.WireDetails
+	actualWire := actual.WireDetails
+	if actualWire == nil {
```
smaye81, line 542: Is it ok to bail here if just `actualWire` is nil? There could potentially be an issue where nothing is set into `actualWire` but the test case is expecting values. In that case, I think we'd miss that failure right?

internal/app/connectconformance/results.go
```diff
@@ -547,3 +535,118 @@ func indent(s string) string {
...
+		}
+		if actualWire.ActualGrpcwebTrailers != nil {
+			errs = append(errs, checkGRPCWebTrailers(actualWire.GetActualGrpcwebTrailers(), actual.ResponseTrailers)...)
+		}
+	}
+
+	return errs
+}
+
+func checkGRPCWebTrailers(rawGRPCWebTrailers string, reportedTrailers []*conformancev1.Header) multiErrors {
```
smaye81, line 594: Might be good to have unit tests for this. It seems sufficiently complex enough that we'd want to test it.

internal/app/connectconformance/results.go
```diff
@@ -547,3 +535,118 @@ func indent(s string) string {
...
+		// gRPC spec. But "net/http" doesn't provide that level of visibility into the
+		// underlying HTTP/2 stream frames. Luckily, the grpc-go reference client DOES
+		// check this. So we can still catch this condition in a misbehaving server.
+
+	case conformancev1.Protocol_PROTOCOL_GRPC_WEB:
+		if len(actualWire.ActualHttpTrailers) > 0 {
+			errs = append(errs, fmt.Errorf("grpc-web response included %d HTTP trailers but should not have any", len(actualWire.ActualHttpTrailers)))
+		}
+		if actualWire.ActualGrpcwebTrailers != nil {
+			errs = append(errs, checkGRPCWebTrailers(actualWire.GetActualGrpcwebTrailers(), actual.ResponseTrailers)...)
```
smaye81, line 587: Does any of the validation being done in `checkGRPCWebTrailers` apply to gRPC trailers also? I know there are some differences, but was just curious why we validate them for gRPC-web but don't validate anything for gRPC.

internal/app/referenceclient/tracer.go
```diff
@@ -117,53 +118,59 @@ func buildWireDetails(ctx context.Context) (*v1.WireDetails, error) {
...
-	} else if strings.HasPrefix(contentType, "application/connect+") {
-		// If this is a streaming request, then look through the trace events
+	case strings.HasPrefix(contentType, "application/connect+"):
+		// If this is a streaming Connect request, then look through the trace events
 		// for the ResponseBodyEndStream event and parse its content into an
 		// endStreamError to see if there are any error details.
-		type endStreamError struct {
+		endStreamContent, ok := getBodyEndStream(trace)
+		if !ok {
+			break
```
smaye81, line 147: I was originally thinking we should report an error here if no `EndStream` message is found for a streaming response. But I guess that would be caught by a test failure with a missing `ConnectErrorRaw` in the response?

internal/app/connectconformance/results.go
```diff
@@ -547,3 +535,118 @@ func indent(s string) string {
...
+
+	if expectedWire.GetActualStatusCode() != 0 {
+		if diff := cmp.Diff(expectedWire.ActualStatusCode, actualWire.ActualStatusCode, protocmp.Transform()); diff != "" {
+			errs = append(errs, fmt.Errorf("actual HTTP status code does not match: - wanted, + got\n%s", diff))
+		}
+	}
+
+	switch testCaseReq.Protocol {
+	case conformancev1.Protocol_PROTOCOL_CONNECT:
+		if len(actualWire.ActualHttpTrailers) > 0 {
```
smaye81, line 555: Maybe a dumb nit, but should we document or validate anywhere that `expected.ActualHttpTrailers` are unused. In other words, if you were to put them in a test yaml file in the `expectedResponse` nothing is actually looking at them. Just a thought, no biggie though.

internal/app/connectconformance/results.go
jhump: Only the reference client populates it. So it will be nil when running tests in client mode (i.e. testing some other client), and that's definitely okay, even if the test case has an "expected" value.

Also, the only thing we should ever put in a test case in here is the status code. Everything else is indirectly checked by the reference client already (since client already processes the connect errors, trailers, etc). So _if_ they are here, then we check them with extra strictness.

Part of me was actually wondering if we should get rid of the wire details and just do all of these checks directly inside the reference client (instead of here in the test runner assertions), and have the client send back any errors via the feedback field. If we did that, the only field we need in the response is the actual status code...

internal/app/connectconformance/results.go
jhump: For gRPC, "net/http" already does all of the validation necessary. They are actual HTTP trailers, and "net/http" is responsible for processing them and providing them in the form of a `http.Header` on the response object.

For gRPC-web, this data is in a chunk of the body, so the application has to process it. And connect-go is a little lenient for these -- which is fine (even preferable) to adhere to the robustness principle -- "be liberal about what you accept but conservative about what you produce". So these are extra checks, just to be more persnickety about the on-the-wire form of gRPC-web trailers.

internal/app/referenceclient/tracer.go
jhump: Correct. Also, we don't expect one for things like cancellations or timeouts. So we'd have to embed way too much logic to get the check right, and the reference client, by virtue of implementing the RPC protocols, is already enforcing the things that need to be enforced. This code is just to scrape add'l on-the-wire details so we can do add'l validation (and be a little more strict than connect-go).

internal/app/connectconformance/results.go
jhump: Yeah, they should not be set in an expected response. See last paragraph [here](https://github.com/connectrpc/conformance/pull/780#discussion_r1483557941). WDYT, should we change up the protos to make it more clear?

smaye81: (modulo the status code which you mentioned above is the only thing we'd keep around in the response)

jhump: Ok. See #784

internal/app/connectconformance/results.go
jhump: You're right. I was being lazy programmer. I'll add tests.

jhump:
@smaye81, okay, I've got this rebased now that all of the wire detail checks are just in the reference client.

I updated this to use the same "printer" abstraction that the reference server uses for logging issues, but instead of getting written to stderr, these get recorded into a `[]string`.

I've also added a test with numerous cases, to verify the logic. And, admittedly, that logic got more complicated so that (1) it can handle different kinds of line endings and complain about them, (2) it can correctly handle "obsolete line folding" described in the spec and complain about it, and (3) validate the field name (in previous iteration, I somehow missed that, and it was only checking the value).

internal/app/referenceclient/wire_details.go
```diff
@@ -0,0 +1,321 @@
...
+
+func examineConnectError(_ json.RawMessage, _ internal.Printer) {
+	// TODO
+}
+
+func examineConnectEndStream(_ json.RawMessage, _ internal.Printer) {
+	// TODO
+}
+
+func examineGRPCEndStream(endStream string, printer internal.Printer) {
```
smaye81, line 201: Holy cow, haha, this is great.  This almost feels like it could be its own package to help others validate end streams.

smaye81 (approved):

jhump:
@smaye81, I just pushed one more commit. Some stuff in that commit were a few small renames and additional doc comments in the wire details stuff, in the hopes of making things a little more clearly named and easier to read.

But I also had to change the capture stuff from using an atomic pointer to using a channel to await the value being available. This is because, sadly, it can't be helped that sometimes there will be some raciness between the goroutine executing the HTTP operation and the RPC caller's goroutine for the timeout cases (only for client- and bidi-stream RPCs; for other kinds of RPCs, the HTTP operation is issued on the _same_ goroutine as the RPC caller).

See 6386c4c46e8b933584001f5378c5216526ef210d.

PR #784, opened 2024-02-08
Remove wire details from proto model

jhump:
As it turns out, the wire details can all be internal implementation details in the reference client. So instead of sending any of that back, the reference client will analyze the details and send back issues in the "feedback" field.

We do still need to send back the HTTP status code though.

Aside: this also adds a `.gitattributes` file so that generated TS/JS files will be collapsed by default in future diffs.


smaye81 (approved):
I like this a lot better 👍

PR #786, opened 2024-02-09
Rename tests and test files

smaye81:
This renames all tests and some test files to adhere to a loose naming convention:

* Any test file that is specific to a protocol should be prefixed with that protocol (i.e. `connect_idempotency.yaml`)
* If a test file contains tests for multiple types of streams, the test name should be prefixed with the stream type. i.e. `{stream type}/{optional description}`
  * Note that in the case of bidi, the convention should be `{stream type}/{[full | half] duplex}/{optional description}`

This allows a lot of flexibility over ignoring tests in `known_failing` files, as well as targeting specific types of tests from the command line (i.e. `server-stream/*` to run all server stream tests).


internal/app/connectconformance/testsuites/data/basic.yaml
```diff
@@ -172,7 +174,7 @@ testCases:
     - "@type": type.googleapis.com/connectrpc.conformance.v1.BidiStreamRequest
 # Misc Tests ------------------------------------------------------------------
 - request:
-    testName: unimplemented
+    testName: unary/unimplemented
```
jhump, line 177: I don't think this one needs a "unary" prefix since the stream type is really irrelevant to what it's testing.

internal/app/connectconformance/testsuites/data/connect_client_code_to_http_code.yaml
```diff
@@ -1,12 +1,12 @@
...
+# according to the protocol.
 mode: TEST_MODE_CLIENT
 relevantProtocols:
   - PROTOCOL_CONNECT
-# These tests verify that an explicit Connect code maps to the correct HTTP code
-# according to the protocol.
 testCases:
 - request:
-    testName: unary canceled mapping
+    testName: unary/canceled
```
jhump, line 9: Same here: these are testing HTTP code mapping and all of them are unary. In fact, it is only the unary connect protocol that even uses HTTP status codes this way, so the "unary/" prefix is redundant in that regard as well. You might even rename the test suite and file to  "Connect Unary to HTTP Code Mapping".

internal/app/connectconformance/testsuites/data/connect_client_http_to_connect_code.yaml
```diff
@@ -1,15 +1,15 @@
...
 # Connect code. These tests cases verify that mapping by forcing the server to
 # return a specified HTTP code and then test whether the client correct returns
 # the required Connect code.
+mode: TEST_MODE_CLIENT
+relevantProtocols:
+  - PROTOCOL_CONNECT
 testCases:
 - request:
-    testName: bad request
+    testName: unary/bad-request
```
jhump, line 12: Ditto again. Everything in here is a unary RPC, so the "unary/" prefix seems like a bit more ceremony than is needed.

internal/app/connectconformance/testsuites/data/connect_idempotency.yaml
```diff
@@ -12,7 +12,7 @@ relevantCodecs:
   - CODEC_PROTO
 testCases:
 - request:
-    testName: idempotent unary success proto
+    testName: unary/success
```
jhump, line 15: Ditto. Only unary calls can use GET, so this feels redundant. Also, this isn't really testing idempotency so much as it is testing the GET HTTP method. So maybe rename the test suite and file to "Connect with GET"

internal/app/connectconformance/testsuites/data/tls-client-certs.yaml
```diff
@@ -4,7 +4,7 @@ reliesOnTlsClientCerts: true
 # This just does the basics with a client-cert, instead of running every test case with them.
 testCases:
   - request:
-      testName: client stream success
+      testName: client-stream/success
```
jhump, line 7: I don't think the "success" suffix is useful since these really all would be basic success tests. Also, maybe a TODO to add unary and other stream type test cases here.

testing/grpcwebclient/known_failing.txt
```diff
@@ -2,21 +2,18 @@
...
-gRPC Web Client/**/unary trailers in body, duplicates
-gRPC Web Client/**/server stream trailers in body, duplicates
-gRPC Web Client/**/unary trailers in body, mixed case
-gRPC Web Client/**/server stream trailers in body, mixed case
+Duplicate Metadata/**/unary/**/success
+Duplicate Metadata/**/server-stream/**
+gRPC Web Client/**/unary/trailers-in-body-duplicates
+gRPC Web Client/**/server-stream/trailers-in-body-duplicates
+gRPC Web Client/**/unary/trailers-in-body-mixed-case
+gRPC Web Client/**/server-stream/trailers-in-body-mixed-case
```
jhump, line 19: I think these can be further flattened now:
```suggestion
gRPC Web Client/**/trailers-in-body-duplicates
gRPC Web Client/**/trailers-in-body-mixed-case
```

jhump (approved):
Some suggestions to further simplify: I don't think it's necessary to state the stream type for every test case since, for many tests, the stream type is not relevant or is implied by the test suite (for example, Connect GET can only ever work with unary). Other than that LGTM.

internal/app/connectconformance/testsuites/data/connect_client_code_to_http_code.yaml
smaye81: There are actually other tests for all the other stream types in this file too that test a `200` response on stream errors. I think the convention of 'if the file contains multiple stream types, prefix the test name with the stream type' is probably a good one.  WDYT?

jhump: My bad. When I scanned through the diff, all I saw was "unary". But I now see that, while most are unary, there are a handful of others below. So disregard this comment.

PR #790, opened 2024-02-13
Add strict checks to reference client to validate a server's connect error and end stream JSON

jhump:
This makes the checks significantly more strict than the base connect-go implementation, in order to report a number of possible issues in server JSON responses:
1. This will detect duplicate keys in JSON objects/maps. Without an explicit check, "encoding/json" ignores duplicate keys, letting later values for the same key overwrite earlier ones. Sadly, there's no configuration to set this uniformly, so it had to be done by overriding the `UnmarshalJSON` behavior of each type to explicitly check. 😦 
2. This ensures that all keys use the correct case. Sadly, even with explicit field names in struct tags, "encoding/json" is still lenient and uses a case-insensitive match for field names. 😢 
3. This double-checks the JSON type of each value. There are a few ways in which "encoding/json" sometimes allows undesired types -- mainly with the use of a JSON null for map, slice, and pointer fields. So this performs an extra check of the actual types in the raw JSON.
4. This makes sure that the keys and values in the end-stream "metadata" map are valid HTTP header names and values.
5. This also makes sure that if a "debug" key is present in an error detail, that the message it describes actually matches the message described in the binary data of the "value" key.

### DIFF
--- a/internal/app/referenceclient/wire_details.go
+++ b/internal/app/referenceclient/wire_details.go
@@ -237,28 +237,8 @@ func getBodyEndStream(trace tracer.Trace) (string, bool) {
 
 func examineConnectError(errJSON json.RawMessage, printer internal.Printer) {
 	var connErr *connectError
-	if err := json.Unmarshal(errJSON, &connErr); err != nil {
-		printer.Printf("connect error JSON is malformed: %v", err)
-		return
-	}
-	if connErr == nil {
-		printer.Printf("connect error JSON: expecting an object but got <nil>")
-		return
-	}
-	// Extra checks, since encoding/json above is lenient.
-	if _, err := checkNoDuplicateKeys("", json.NewDecoder(bytes.NewReader(errJSON))); err != nil {
-		printer.Printf("connect error JSON: %v", err)
-		return
-	}
-	var asAny map[string]any
-	if err := json.Unmarshal(errJSON, &asAny); err != nil {
-		// note: since above unmarshal step succeeded, this should never fail
-		printer.Printf("connect error JSON is malformed: %v", err)
-		return
-	}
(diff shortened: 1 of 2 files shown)


jhump:
@smaye81, this one's pretty big. Sorry about that. Though the test cases are a pretty big chunk of the lines changed. But there's also a non-trivial amount of logic to closely scrutinize the JSON data and complain about it.

internal/app/referenceclient/wire_details.go
```diff
@@ -206,12 +235,255 @@ func getBodyEndStream(trace tracer.Trace) (string, bool) {
...
+	if _, err := checkNoDuplicateKeys("", json.NewDecoder(bytes.NewReader(endStreamJSON))); err != nil {
+		printer.Printf("connect end stream JSON: %v", err)
+		return
+	}
+	var asAny map[string]any
+	if err := json.Unmarshal(endStreamJSON, &asAny); err != nil {
+		// note: since above unmarshal step succeeded, this should never fail
+		printer.Printf("connect end stream JSON is malformed: %v", err)
+		return
+	}
```
smaye81, line 444: Is it worth moving out this logic into its own function? I noticed that `examineConnectError` and `examineConnectErrorDetail` perform very similar logic (albeit with some minor variations). All 3 seem to do a series of validation steps against a `json.RawMessage` and then use an `internal.Printer` to report an error. It may not even be worth it, but just an observation.

smaye81 (approved):

internal/app/referenceclient/wire_details.go
jhump: Interesting point. I looked and was able to consolidate this into an `examineJSON` helper function. Just pushed a change with that.

PR #791, opened 2024-02-13
Add docs for writing test cases

smaye81:
This adds some documentation for writing test cases, including:

* Location of files
* Test and suite naming conventions
* Directives for suites as well as links to the BSR proto file and other helpful examples.


jhump (commented):
This needs a link to the schema for the YAML files, so users know where to look for more relevant docs.

We might also consider replicating some of the content from the proto comments here so this page can provide a little more "how to" vs. requiring the user to read comments for every field in the test suite schema.

In particular, we should re-iterate the fields in the schema that should _not_ be put into the test case -- like the host name and port, TLS details, and even standard/protocol-oriented request and response headers and trailers (only custom headers belong in the test case).

I think it would also be useful to show a (small-ish) example in full, and maybe also describe some of the basics, like the HTTP versions, protocols, compression algorithms, and stream types.

And in the section mentioning how the test can be constrained with the "relevant ..." fields, you might also  describe briefly how _permutations_ of each test case will be computed, based on the features supported by the client or server under test and the relevant properties of the test suite.

Though perhaps the above -- a section on test case permutations -- belongs in the page about configuration, since it's most relevant to someone configuring and running the tests for their client or server implementation. I guess ideally it could be mentioned here and then linked over to more information in the other page. So maybe we can come back to it and that breadcrumb once the target doc is written...

smaye81:
@jhump expanded the docs a bit. lmk what you think.

internal/app/connectconformance/test_case_library.go
```diff
@@ -441,6 +453,7 @@ func populateExpectedResponse(testCase *conformancev1.TestCase) error {
 	// message in this situation, where the response data value is some fixed string (such as "no response definition")
 	// and whose request info will still be present, but we expect it to indicate zero request messages.
 	if len(testCase.Request.RequestMessages) == 0 {
+		testCase.ExpectedResponse = &conformancev1.ClientResponseResult{}
```
smaye81, line 456: This was causing a panic if you omitted `requestMessages` from a test case. Since `ExpectedResponse` was nil here, once we returned, the `assert` function in results.go was blowing up on line ~170.

docs/authoring_test_cases.md
```diff
@@ -1,11 +1,83 @@
...
+Once the above are specified, you can then define your request. For a full list of fields to specify in the request,
+see the [`ClientCompatRequest`][client-compat-proto] message in the Conformance Protobuf definitions.
+
+ > [!IMPORTANT]  
+ > The `ClientCompatRequest` message contains some fields that should _not_ be specified in test cases.
+ > These fields include:
+ > * Fields 1 through 8 in the message definition. These fields are automatically populated by the test runner.
+ >   If a test is specific to one of these values, it should instead be indicated in the directives for the test suite
+ >   itself.
+ > * Field 20 (`raw_request`). This field is only used by the reference client for sending anomalous requests.
```
jhump, line 79: This doesn't really belong in this list. Users will need to specify this in test cases that need to test exceptional wire-format cases.

In fact, it's probably worth a section in here about the use of raw requests (which are only handled by the reference client and should only appear in files with "server" mode) and raw responses (vice versa, only for files with "client" mode, handled by the reference server).

docs/authoring_test_cases.md
```diff
@@ -1,11 +1,83 @@
...
+  to `false`.
+
+## Test cases
+
+Test cases are specified in the `testCases` property of the suite. Each test case starts with the `request` property 
+which defines the request which will be sent to a client during the test run. Each `request` must specify the following
+fields:
+
+* `testName` - For naming conventions, see [below](#naming-conventions).
+* `service` - The fully-qualified path of the service this test will interact with, typically the `ConformanceService`.
```
jhump, line 65:
```suggestion
* `service` - The fully-qualified name of the service this test will interact with. For now, this is always `connectrpc.conformance.v1.ConformanceService`.
```

docs/authoring_test_cases.md
```diff
@@ -1,11 +1,83 @@
...
+* `method` - This is a string specifying the method on `service` that will be called.
+* `streamType` - One of `STREAM_TYPE_UNARY`, `STREAM_TYPE_CLIENT_STREAM`, `STREAM_TYPE_SERVER_STREAM`, 
+  `STREAM_TYPE_HALF_DUPLEX_BIDI_STREAM`, or `STREAM_TYPE_FULL_DUPLEX_BIDI_STREAM`.
+
+Once the above are specified, you can then define your request. For a full list of fields to specify in the request,
+see the [`ClientCompatRequest`][client-compat-proto] message in the Conformance Protobuf definitions.
```
jhump, line 71: I think it will be more helpful to link to the actual message description in the generated docs instead of linking to the source of the containing file
https://buf.build/connectrpc/conformance/docs/main:connectrpc.conformance.v1#connectrpc.conformance.v1.ClientCompatRequest

internal/app/connectconformance/test_case_library.go
```diff
@@ -174,6 +174,18 @@ func (lib *testCaseLibrary) expandSuite(suite *conformancev1.TestSuite, configCa
 
 func (lib *testCaseLibrary) expandCases(cfgCase configCase, namePrefix []string, testCases []*conformancev1.TestCase) error {
 	for _, testCase := range testCases {
+		if testCase.Request.TestName == "" {
```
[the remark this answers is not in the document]
jhump: You could provide a little bit more context for these errors if you change the loop vars from `_, testCase` to `i, testCase`:
```go
return fmt.Errorf("test case #%d: test case has no name", i+1)
```

jhump (commented):
Looks really good. I left a few more comments, but I think this is close.

smaye81:
@jhump Updated w/ latest feedback. Lmk what you think.

jhump (approved):
LG!

PR #792, opened 2024-02-16
Add additional docs to the proto files

smaye81:
This expands on documentation in the proto files and also adds documentation to messages where none existed.


proto/connectrpc/conformance/v1/config.proto
```diff
@@ -32,42 +32,61 @@ message Config {
...
+  // Whether trailers are supported.
   // If absent, true is assumed. If false, implies that gRPC protocol is not allowed.
   optional bool supports_trailers = 9;
+  // Whether half duplex bidi streams are supported over HTTP/1.1.
   // If absent, false is assumed.
   optional bool supports_half_duplex_bidi_over_http1 = 10;
+  // Whether Connect via GET is supported.
   // If absent, true is assumed.
   optional bool supports_connect_get = 11;
+  // Whether the Connect version header (Connect-Protocol-Version) is required.
```
smaye81, line 80: Does this field also apply to the Connect version query param? If so, should we consider renaming?

jhump: I don't think we actually use it at all. I had some ideas early on (when I added this field) about writing tests that verified this -- like a test case that would omit it and be expected to fail if this was set to true (and another case, for when this was false, that would omit and expect to succeed). But they never to came to pass.

We should probably instead remove it entirely. It might be nice to add it as another check to the reference server -- so even though it's technically optional in the protocol spec, we'd verify that conformant client implementations always include it (to maximize inter-op, just in case a server is configured to require it).

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -57,18 +65,32 @@ message ClientCompatRequest {
...
   // when making the request.
   bool use_get_http_method = 14;
+  // Any request headers that should be sent as part of the request.
   repeated Header request_headers = 15;
-  // There will be exactly one for unary and server-stream methods.
+  // The actual request messages that will sent to the server.
+  // The type URL for all entries should be equal to the request type of the
+  // method.
+  // There will be exactly one for unary and server-stream methods and 
+  // one-to-many for client- and bidi-stream methods.
```
jhump, line 88:
```suggestion
  // There must be exactly one for unary and server-stream methods but 
  // can be zero or more for client- and bidi-stream methods.
```

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -57,18 +65,32 @@ message ClientCompatRequest {
...
+  // When writing test cases, this is a required field.
   string method = 12;
+  // The stream type of `method` (i.e. Unary, Client-Streaming, Server-Streaming, Full Duplex Bidi, or Half Duplex Bidi).
+  // When writing test cases, this is a required field.
   StreamType stream_type = 13;
   // If protocol indicates Connect and stream type indicates
   // Unary, this instructs the client to use a GET HTTP method
   // when making the request.
   bool use_get_http_method = 14;
+  // Any request headers that should be sent as part of the request.
```
jhump, line 82:
```suggestion
  // Any request headers that should be sent as part of the request.
  // These include only custom header metadata. Headers that are
  // part of the relevant protocol (such as "content-type", etc) should
  // not be stated here.
```

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -57,18 +65,32 @@ message ClientCompatRequest {
...
+  // The type URL for all entries should be equal to the request type of the
+  // method.
+  // There will be exactly one for unary and server-stream methods and 
+  // one-to-many for client- and bidi-stream methods.
   // For client- and bidi-stream methods, all entries will have the
-  // same type URL (which matches the request type of the method).
+  // same type URL.
   repeated google.protobuf.Any request_messages = 16;
+  // The timeout, in milliseconds, for the request. This is equivalent to a
+  // deadline for the request.
```
jhump, line 93:
```suggestion
  // The timeout, in milliseconds, for the request. This is equivalent to a
  // deadline for the request. If unset, there will be no timeout.
```

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -114,41 +136,53 @@ message ClientCompatRequest {
 
 // The outcome of one ClientCompatRequest.
 message ClientCompatResponse {
+  // The test name that this response applies to.
   string test_name = 1;
+  // These fields determine the outcome of the request. Note that an `error`
+  // should only be reported for unexpected internal errors. RPC errors, such
+  // as Connect error codes, should be reported as part of the `response`.
```
jhump, line 143: Another way to say this, which might help clarify, is that any unexpected errors that prevent the client from issuing the RPC and following the instructions implied by the request can be reported as an `error`. These would be errors creating an RPC client from the request parameters or unsupported/illegal values in the request (e.g. a unary request that defines zero or multiple request messages). But once the RPC is issued, any resulting error should instead be encoded in `response`.

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -114,41 +136,53 @@ message ClientCompatRequest {
...
+  // All response headers read from the response.
   repeated Header response_trailers = 4;
-
   // The number of messages that were present in the request but that could not be
   // sent because an error occurred before finishing the upload.
   int32 num_unsent_requests = 5;
-
   // The following field is only set by the reference client. It communicates
   // the underlying HTTP status code of the server's response.
+  // If you are implementing a client-under-test, you may ignore this field.
```
jhump, line 171:
```suggestion
  // If you are implementing a client-under-test, you should ignore this field
  // and leave it unset.
```

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -114,41 +136,53 @@ message ClientCompatRequest {
...
   optional int32 http_status_code = 6;
-
   // This field is used only by the reference client, and it can be used
   // to provide additional feedback about problems observed in the server
   // response or in client processing of the response. If non-empty, the test
   // case is considered failed even if the result above matches all expectations.
+  // If you are implementing a client-under-test, you may ignore this field.
```
jhump, line 177:
```suggestion
  // If you are implementing a client-under-test, you should ignore this field
  // and leave it unset.
```

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -114,41 +136,53 @@ message ClientCompatRequest {
...
   repeated string feedback = 7;
 }
 
 // The client is not able to fulfill the ClientCompatRequest. This may be due
 // to a runtime error or an unexpected internal error such as the requested protocol
 // not being supported. This is completely independent of the actual RPC invocation.
 message ClientErrorResult {
+  // A message describing the error that occurred.
```
jhump, line 185: Maybe mention that this string will be shown to users running conformance tests so should include any relevant details that may help them troubleshoot or remedy the error.

proto/connectrpc/conformance/v1/config.proto
```diff
@@ -32,42 +32,61 @@ message Config {
   repeated ConfigCase exclude_cases = 3;
 }
 
+// Features define the feature set that a client or server support. They are
+// used to determine the tests that are possible to be run in a client-server
+// combination. They are defined in YAML files and are specified as part of the
+// --conf flag to the test runner.
```
jhump, line 38:
```suggestion
// Features define the feature set that a client or server supports. They are
// used to determine the server configurations and test casess that are 
// will be run. They are defined in YAML files and are specified as part of the
// --conf flag to the test runner.
```
Perhaps, though, much of this comment instead belongs on `Config`. After all `Config` is the schema from the `--conf` yaml file, not this message.

proto/connectrpc/conformance/v1/config.proto
```diff
@@ -32,42 +32,61 @@ message Config {
...
+  // Whether the Connect version header (Connect-Protocol-Version) is required.
   // If absent, false is assumed.
   optional bool requires_connect_version_header = 12;
+  // Whether a message receive limit is supported.
   // If absent, true is assumed.
   optional bool supports_message_receive_limit = 13;
 }
 
+// ConfigCase represents a resolved configuration case for a single test based
+// on features specified.
```
jhump, line 89:
```suggestion
// ConfigCase represents a single resolved configuration case. When tests are
// run, the Config and the supported features therein are used to compute all
// of the cases relevant to the implementation under test. These configuration
// cases are then used to select which test cases are applicable.
```

proto/connectrpc/conformance/v1/server_compat.proto
```diff
@@ -33,10 +33,12 @@ import "connectrpc/conformance/v1/config.proto";
...
 // will be started, each with different properties.
 message ServerCompatRequest {
+  // The protocol that will be used.
   Protocol protocol = 1;
+  // The HTTP version that will be used.
   HTTPVersion http_version = 2;
-  // if true, generate a self-signed cert and include it in the
-  // ServerCompatResponse along with the actual port
+  // If true, generate a self-signed cert and include it in the
+  // ServerCompatResponse along with the actual port.
```
jhump, line 41:
```suggestion
  // If true, generate a self-signed cert and include it in the
  // ServerCompatResponse along with the actual port. Clients
  // will be configured to trust this cert when connecting.
  // If false, the server should not use TLS and instead use
  // a plain-text/unencrypted socket.
```

proto/connectrpc/conformance/v1/server_compat.proto
```diff
@@ -63,5 +67,6 @@ message ServerCompatResponse {
 // The server doesn't support the requested protocol, or had a runtime error
 // while starting up.
 message ServerErrorResult {
```
jhump, line 69: Is this even used? I think maybe it's not and I forgot to actually delete this message. I think what actually happens is that the server should simply exit with a non-zero exit code on error and print a relevant error message to stderr.

proto/connectrpc/conformance/v1/service.proto
```diff
@@ -316,43 +318,53 @@ message ConformancePayload {
...
   message ConnectGetInfo {
+    // The query params observed in the request URL.
     repeated Header query_params = 1;
   }
 }
 
 // An error definition used for specifying a desired error response
 message Error {
+  // The error code.
   int32 code = 1;
```
jhump, line 329: We might want to change to use an enum. The existing YAML files will all still work (since enums can be provided via their numeric value in JSON and YAML). But it would allow us to replace the numbers with human-readable enum names.

WDYT?

proto/connectrpc/conformance/v1/service.proto
```diff
@@ -316,43 +318,53 @@ message ConformancePayload {
...
   }
   message ConnectGetInfo {
+    // The query params observed in the request URL.
     repeated Header query_params = 1;
   }
 }
 
 // An error definition used for specifying a desired error response
 message Error {
+  // The error code.
```
jhump, line 328: I would link to relevant docs:
https://connectrpc.com/docs/protocol#error-codes

proto/connectrpc/conformance/v1/service.proto
```diff
@@ -316,43 +318,53 @@ message ConformancePayload {
...
   int32 code = 1;
   // If this value is absent in a test case response definition, the contents of the
   // actual error message will not be checked. This is useful for certain kinds of
   // error conditions where the exact message to be used is not specified, only the
   // code.
   optional string message = 2;
+  // Error details.
```
jhump, line 335:
```suggestion
  // Errors in Connect and gRPC protocols can have arbitrary messages
  // attached to them, which are known as error details.
```

proto/connectrpc/conformance/v1/service.proto
```diff
@@ -316,43 +318,53 @@ message ConformancePayload {
...
   repeated google.protobuf.Any details = 3;
 }
 
 // A tuple of name and values (ASCII) for a header or trailer entry.
 message Header {
+  // Header/trailer name (key).
   string name = 1;
+  // Header/trailer value.
```
jhump, line 343:
```suggestion
  // Header/trailer value. This is repeated to explicitly support headers and
  // trailers where a key is repeated. In such a case, these values must be in
  // the same order as which values appeared in the header or trailer block.
```

proto/connectrpc/conformance/v1/suite.proto
```diff
@@ -25,8 +25,13 @@ import "connectrpc/conformance/v1/config.proto";
...
 // based on whether the client or server under test implements these features.
 message TestSuite {
+  // Test suite name. When writing test suites, this is a required field.
   string name = 1;
-
+  // The mode (client or server) that this test suite applies to. This is used
+  // in conjunction with the `--mode` flag passed to the conformance runner
+  // binary. If the mode on the suite is set to client, the tests will only be
+  // run if `--mode client` is set on the command to the test runner.
+  // Likewise if mode is server.
```
jhump, line 34:
```suggestion
  // Likewise if mode is server. If this is unset, the test case will be run in both modes.
```

proto/connectrpc/conformance/v1/service.proto
smaye81: Cool. I'll add it to this PR. wdyt about enum name?
* `ConnectErrorCode`
* `ErrorCode`
* `Code

proto/connectrpc/conformance/v1/client_compat.proto
```diff
@@ -57,18 +65,32 @@ message ClientCompatRequest {
   // If the server sends anything larger, the client should reject it.
   uint32 message_receive_limit = 10;
 
+  // The fully-qualified name of the service this test will interact with. 
+  // For now, this is always connectrpc.conformance.v1.ConformanceService.
+  // When writing test cases, this is a required field.
   string service = 11;
+  // The method on `service` that will be called.
+  // When writing test cases, this is a required field.
   string method = 12;
```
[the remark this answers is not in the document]
smaye81: Yep, that works. I can add it to this PR.

proto/connectrpc/conformance/v1/config.proto
smaye81: Ah yeah, this isn't used either. Cool to add a TODO to the server checks?

proto/connectrpc/conformance/v1/service.proto
jhump: I think `Code` would be the most clear, since that matches the name used by both connect-go and grpc-go.

proto/connectrpc/conformance/v1/config.proto
jhump: Certainly!

smaye81: Actually in looking at it. It's not set anywhere in a config YAML, but it does indirectly get used in `internal/app/connectconformance/config.go`. Not sure if all of that should be ripped or not?

jhump: Yes, it should be ripped out. But a TODO is fine as well.

smaye81:
@jhump ok, incorporated all feedback except for the `requires_connect_version_header` removal. Doing so caused a chain reaction through a lot of files that I wasn't sure should be done. I can put it up as a separate PR so that it's easier to visualize.

Also, I made `service` and `method` optional and default them if not populated. The non-uniqueness of the `UNARY` stream type makes things a bit more convoluted, but it does allow us to shrink all the test files. I'll let you be the judge. Lmk what you think.

proto/connectrpc/conformance/v1/config.proto
smaye81: Mentioned below. Going to put this in its own PR.

internal/app/connectconformance/test_case_library.go
```diff
@@ -177,18 +178,33 @@ func (lib *testCaseLibrary) expandCases(cfgCase configCase, namePrefix []string,
...
 		if testCase.Request.StreamType != cfgCase.StreamType {
 			continue
 		}
+		// If service isn't specified, default to the Conformance Service.
+		if testCase.Request.GetService() == "" {
+			serviceName := conformancev1connect.ConformanceServiceName
+			testCase.Request.Service = &serviceName
+		}
+		// If method isn't specified, auto-populate based on stream type
+		if testCase.Request.GetMethod() == "" {
```
jhump, line 193: Nit from above comment: we should only fill in a default value for the method if the service name is also blank.

We could even require that we never set a default on either unless both are blank. If we went that far, then the tests for "unimplemented" would have to specify both the service and method name, which I think might be okay and even best for clarity.

internal/app/grpcclient/client.go
```diff
@@ -163,10 +163,10 @@ func invoke(ctx context.Context, req *v1.ClientCompatRequest) (*v1.ClientRespons
 		return nil, err
 	}
 
-	switch req.Service {
+	switch *req.Service {
```
jhump, line 166: This is where the generated accessors are usually used.
```suggestion
	switch req.GetService() {
```

internal/app/grpcclient/impl.go
```diff
@@ -44,7 +44,7 @@ func (i *invoker) Invoke(
 		ctx = deadlineCtx
 		defer cancel()
 	}
-	switch req.Method {
+	switch *req.Method {
```
jhump, line 47: ditto: `req.GetMethod()`

internal/app/referenceclient/client.go
```diff
@@ -329,11 +329,11 @@ func invoke(ctx context.Context, req *v1.ClientCompatRequest, referenceMode bool
 		clientOptions = append(clientOptions, connect.WithReadMaxBytes(int(req.MessageReceiveLimit)))
 	}
 
-	switch req.Service {
+	switch *req.Service {
```
jhump, line 332: here, too

internal/app/referenceclient/impl.go
```diff
@@ -62,7 +62,7 @@ func (i *invoker) Invoke(
 		defer cancel()
 	}
 
-	switch req.Method {
+	switch *req.Method {
```
jhump, line 65: and here

jhump (approved):
Looks awesome!

smaye81:
@jhump incorporated latest feedback in https://github.com/connectrpc/conformance/pull/792/commits/ae01d566b7359d0d3fb883643cd84dd6e90fea33. lmk if it seems ok.

jhump (approved):
Very nice! 💥

PR #794, opened 2024-02-17
Lots more test cases

jhump:
Adds numerous new test cases for verifying edge cases and error conditions for all three protocols.

I think this should round out all of the tests we want for v1.0.0-rc3. So after this lands, we can cut a new release of the conformance stuff and then see what it shakes out in other implementations 🤞.


internal/app/referenceclient/wire_details_test.go
```diff
@@ -609,9 +609,11 @@ func TestExamineGRPCEndStream(t *testing.T) {
 			} else {
 				// When there's feedback, the connect-go client may complain about the end-stream message
 				// and report a different code.
-				assert.True(t, connect.CodeOf(err) == connect.CodeAlreadyExists ||
-					connect.CodeOf(err) == connect.CodeInternal,
-					"unexpected error: %v", err)
+				assert.Contains(t,
+					[]connect.Code{connect.CodeAlreadyExists, connect.CodeInternal, connect.CodeUnknown},
```
jhump, line 613: With fixes in latest connect-go, some of these scenarios can now result in "unknown" code.

internal/tracer/reader.go
```diff
@@ -116,6 +118,9 @@ type dataTracer struct {
 }
 
 func (d *dataTracer) trace(data []byte) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
jhump, line 122: I happened to trip the race detector when running tests at one point, perhaps because of the add'l test cases added to the `TestRun` test in `connectconformance_test.go`.

I forgot that the body of an `http.Request` needs to be thread-safe to at least support a call to `Close` being concurrent with a call to `Read`. And that was not the case before, but now it is safe, due to this locking.

internal/app/connectconformance/connectconformance_test.go
```diff
@@ -77,7 +101,7 @@ func TestRun(t *testing.T) {
 		allSuites,
 		logger,
 		logger,
-		&Flags{Verbose: true, MaxServers: 2, Parallelism: 4},
+		&Flags{Verbose: true, MaxServers: 2, Parallelism: 4, ServerBind: "127.0.0.1"},
```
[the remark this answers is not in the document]
smaye81: I saw this too. So glad you fixed it.

smaye81:
lgtm - although you may want to remove the `service` and `method` values in the test files if we decide to land https://github.com/connectrpc/conformance/pull/792. 😬 

Also, I wonder if maybe we should create tasks or issues, etc. for a lot of the TODOs we have, at least as they apply to skipped tests, known failing, etc.?

smaye81 (approved):

jhump:
> although you may want to remove the service and method values in the test files if we decide to land #792. 😬

Done! I also pushed a commit that changes the `code` values in the YAML files to use the new code names instead of just numbers. In some of the gRPC cases, I left them as numbers since it's clear they are echoing back the value in a `grpc-status` trailer (which just uses numbers, not names). But I changed all of the others.

I also found a commented-out test case that I forgot to uncommment, that passes with the latest connect-go, in [`connect_client_http_to_connect_code.yaml`](https://github.com/connectrpc/conformance/pull/794/commits/d94b32bbf6d64c08168b65eb881f45aefbbdf578#diff-571a3ce2d2483038775cb689a79376dea9a0104b8d6167702f72780d34997fb8L76-L89).

PR #799, opened 2024-02-20
Make all import aliases consistent

smaye81:
This makes all import aliases for the Conformance protos consistent. Previously, there was a mix of `v1` and `conformancev1` and this ensures they're all `conformancev1`.


jhump (approved):
It might easiest to land #792 first, and then rebase this. I saw some new references to `v1.` prefixes in that PR with some of the code changes.

PR #805, opened 2024-02-23
Tighten up when response can combine headers and trailers; add check for unsent requests

jhump:
Doh! I made all these changes a while ago and was waiting on the connect-go release (since the tests don't pass without some of the latest changes/fixes).

I had intended to include this in rc3. Oh well... maybe in an rc4 or maybe in a final v1.0.0.

### DIFF
--- a/internal/app/referenceclient/impl.go
+++ b/internal/app/referenceclient/impl.go
@@ -358,7 +358,7 @@ func (i *invoker) clientStream(
 	resp, err := stream.CloseAndReceive()
 	if err != nil {
 		// If an error was returned, first convert it to a Connect error
-		// so that we can get the headers from the Meta property. Then,
+		// so that we can get the trailers from the Meta property. Then,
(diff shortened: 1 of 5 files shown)


internal/app/referenceclient/impl.go
```diff
@@ -360,7 +361,7 @@ func (i *invoker) clientStream(
 		// so that we can get the headers from the Meta property. Then,
```
smaye81, line 361: The comment should also be `trailers` right (like the change above)?

smaye81 (approved):

internal/app/referenceclient/impl.go
jhump: Yes, good catch!

jhump:
```suggestion
		// so that we can get the trailers from the Meta property. Then,
```

PR #808, opened 2024-02-26
Add docs for implementing a server

smaye81:
This adds documentation for how to implement a server for conformance tests.


smaye81:
@jhump This probably still needs some work, but wanted to get some concrete feedback. All the content is there, but let me know if there's anything I'm missing and if you think the pseudocode examples make sense.

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+* Implementing the ConformanceService endpoints.
+* Making an executable file that will read messages from `stdin` and write messages to `stdout`.
+* Defining any configuration for what your server supports. For more information on how to do this, see the [Configuration Files](./configuring_and_running_tests.md#configuration-files) section in the [Configuring and Running Tests documentation](./configuring_and_running_tests.md).
+
+## Process
+
+The basic process for working with servers in the conformance suite will work as follows:
+
+* The conformance runner will pass a `ServerCompatRequest` message via `stdin`. This message will contain all the details necessary
```
jhump, line 13: We need a little more clarity on what it means to "pass a message via stdin". In particular, this should describe how the program will need to read four bytes for the 32-bit (network byte order) size of the message, then read that many bytes, and then de-serialize those bytes into this message type.

Same for the "write a message to stdout" below.

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
+# Testing Servers
+
+Testing servers with the conformance runner involves the following steps:
+
+* Implementing the ConformanceService endpoints.
+* Making an executable file that will read messages from `stdin` and write messages to `stdout`.
+* Defining any configuration for what your server supports. For more information on how to do this, see the [Configuration Files](./configuring_and_running_tests.md#configuration-files) section in the [Configuring and Running Tests documentation](./configuring_and_running_tests.md).
```
jhump, line 7: I don't think we need two links here. How about this?
```suggestion
* Defining any configuration for what your server supports. For more information on how to do this, see the docs for [configuring and running tests](./configuring_and_running_tests.md#configuration-files).
```

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+The basic process for working with servers in the conformance suite will work as follows:
+
+* The conformance runner will pass a `ServerCompatRequest` message via `stdin`. This message will contain all the details necessary
+  for your implementation to start its server-under-test. Note that the request does not specify a port to listen on.
+  Implementations should instead pick an available ephemeral port according to their OS and return that value in the response.
+
+  Fields in the request are:
+
+  * `protocol` specifies what protocol will be used to run the tests. 
+  * `http_version` specifies which HTTP version will be used.
```
jhump, line 20: Maybe we could clarify here, too. Some HTTP server libraries may automatically provide support for both HTTP 1.1 and HTTP/2, in which case it is fine for the server program to always support both and ignore this field.

So both of these fields really mean "support _at least_ this protocol and version; fine to also support others".

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+The basic process for working with servers in the conformance suite will work as follows:
+
+* The conformance runner will pass a `ServerCompatRequest` message via `stdin`. This message will contain all the details necessary
+  for your implementation to start its server-under-test. Note that the request does not specify a port to listen on.
+  Implementations should instead pick an available ephemeral port according to their OS and return that value in the response.
+
+  Fields in the request are:
+
+  * `protocol` specifies what protocol will be used to run the tests.
```
jhump, line 19: If the server implementation under test always supports multiple protocols, then it is fine to ignore this field. But this is presented in case the server needs to do anything extra to support particular protocols.

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+* The conformance runner will pass a `ServerCompatRequest` message via `stdin`. This message will contain all the details necessary
+  for your implementation to start its server-under-test. Note that the request does not specify a port to listen on.
+  Implementations should instead pick an available ephemeral port according to their OS and return that value in the response.
+
+  Fields in the request are:
+
+  * `protocol` specifies what protocol will be used to run the tests. 
+  * `http_version` specifies which HTTP version will be used.
+  * `use_tls` specifies whether your server should generate a self-signed certificate. Clients will be configured to trust
```
jhump, line 21: FWIW, it is not _required_ for the cert to be self-signed. It can be any TLS certificate where the subject matches the host name that gets sent back in the `ServerCompactResponse`.

But we explicitly _allow_ self-signed (and "localhost" as the subject) to make it easier to implement the server w/out having to worry about certificate issuance, managing/distributing certificate secrets, etc.

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+  * `client_tls_cert` represents a PEM-encoded certificate that, if provided, clients will use to authenticate themselves. 
+     If specified, servers should require that clients provide certificates and they should ensure the presented certificate is valid.
+     Note that this will always be empty if `use_tls` is `false`.
+  * `message_receive_limit` specifies the maximum size in bytes for a message. If this value is non-zero, servers should reject
+     any message from a client that is larger than the size indicated.
+
+* Once the server is started, your implementation should then write a `ServerCompatResponse` message to `stdout`. This will
+  provide the conformance runner with details about your running server.
+
+  * `host` should be set with the host where your server is running.
```
jhump, line 33: This should usually be "127.0.0.1", unless the program actually starts a remote server to which the client should connect.

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+     Note that this will always be empty if `use_tls` is `false`.
+  * `message_receive_limit` specifies the maximum size in bytes for a message. If this value is non-zero, servers should reject
+     any message from a client that is larger than the size indicated.
+
+* Once the server is started, your implementation should then write a `ServerCompatResponse` message to `stdout`. This will
+  provide the conformance runner with details about your running server.
+
+  * `host` should be set with the host where your server is running. 
+  * `port` should be set with the port number where your server is listening.
+  * `pem_cert` should contain the self-signed certificate generated by your server if `use_tls` was set to `true`. Clients
```
jhump, line 35:
```suggestion
  * `pem_cert` should contain the self-signed certificate generated by your server if `use_tls` was set to `true`, in PEM format. Clients
```

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+**Pseudocode**
+
+```text
+read the request
+
+capture any request headers and the actual request body as request info
+
+if a response definition was specified then
+  if response definition specifies valid response data then
```
jhump, line 78: Maybe we can tweak this condition and then swap the two branches? I think it would be more clear if it were more like "if response definition specifies an error then"

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+    build a conformance payload
+
+    set the request info and response data into the conformance payload
+
+  else
+    build an error with the specified error and set the request info into the error details
```
jhump, line 85: I think we might want a sentence or paragraph (outside the pseudo-code) about error details and storing the request info as an error detail -- and reference the precise proto message type (and even link to its docs in the BSR).

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+  set any response headers or trailers indicated in the response definition
+
+else
```
jhump, line 89: I think the conditionals could be simplified to something like:
```
if response definition specifies an error then

   // do error

else

    build a conformance payload with the request info
    if a response definition exists then set the payload's data

end if
```
(For that last line: I wonder if it would be easier to read with explicit "end" lines or even curly braces. As is, it could be a little confusing for anyone but a Python developer 😛. WDYT?)

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+  build a conformance payload
+
+  set the request info into the conformance payload
+
+sleep for any specified response delay
+
+return the response with conformance payload or the error depending on which was specified
```
jhump, line 96: This probably depends on the programming language. In some languages, errors are thrown/raised instead of returned. Maybe tweak the verb?
```suggestion
return the response with conformance payload or raise the error, depending on which was specified
```

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+* the server should echo back all request properties, including all request
+  messages in the order they were received, in the first response. Subsequent
+  responses should only include the message data in the data field.
+  
+* if the `response_delay_ms` duration is specified, the server should wait that
+  long in between sending each response message.
+
+**Pseudocode**
+
+```text
```
jhump, line 271: I think this might be easier to follow if we split it into two -- pseudo-code for the full-duplex logic and separate pseudo-code for the half-duplex logic.

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+The only major difference is that this endpoint should be invoked via an HTTP `GET`. As a result, there is no request body
+so the endpoint should read any query params and set them accordingly in the `connect_get_info` field of the 
+returned `ConformancePayload`. This RPC is the only one that sets the `connect_get_info` field.
+
+For the full documentation on implementing the `IdempotentUnary` endpoint, click [here][idempotentunary].
+
+### Unimplemented
+
+The `Unimplemented` endpoint is also a unary operation, but contrary to the above unary endpoints, the implementation
+of `Unimplemented` should simply return an `unimplemented` error. It is not necessary to echo back any request information or
```
jhump, line 116: Actually, it should not be implemented at all. This tests that the server framework can automatically return the "unimplemented" error for unimplemented methods. Part of the gRPC spec is that adding methods to a service is a safe/compatible change, and that server frameworks should handle this by automatically responding to methods that are missing from the actual server handler implementation. (In connect-go, this actually gets handled by the generated code, in the `Unimplemented...` generated server handler struct, instead of elsewhere in the connect-go runtime code.)

docs/testing_servers.md
```diff
@@ -0,0 +1,340 @@
...
+
+For the full documentation on implementing the `Unary` endpoint, click [here][unary].
+
+
+### IdempotentUnary
+
+The `IdempotentUnary` endpoint is also a unary operation. It accepts a request of type  `IdempotentUnaryRequest` and returns
+a single response of type `IdempotentUnaryResponse`. It should be handled in mostly the same way as `Unary`.
+
+The only major difference is that this endpoint should be invoked via an HTTP `GET`. As a result, there is no request body
```
jhump, line 107: I think it's worth clarifying this a little more. In particular that implementations that only support gRPC or gRPC-Web do not need to implement it. Also, Connect implementations that do not support GET requests also don't need to implement it.

smaye81:
@jhump Addressed feedback. Lmk what you think.

jhump (approved):
Ship it!

PR #811, opened 2024-02-27
In referenceclient, inspecting wire details needs to support compressed error bodies

### LINKED ISSUE #809: Runner should allow compressed error responses
Runner should allow compressed error responses. Runner currently fails if the server returns a compressed error response.

jhump:
This also adds test cases -- both for the reference client's wire details functionality and also conformance test cases, to verify correct support in RPC client implementations for compressed errors.

There's also a small unrelated change so we don't bother validating the "debug" key in JSON error detail if the "value" bytes aren't correctly decoded. (Otherwise, it just reports spurious nonsense.)

Fixes #809


smaye81 (approved):

jhump:
cc @srikrsna-buf

PR #815, opened 2024-03-01
Prevent errant 'content-length: 0' request header from being printed

jhump:
I happened to notice in a trace output a "Content-Length: 0" that was _not_ possible. It turns out, the logic here was incorrectly deciding to add it to the trace output.

The `http.Request.ContentLength` field is actually interpreted differently between client and server requests (those created by a client, to send via a `http.RoundTripper`, and those created by a server to provide to an `http.Handler`). It turns out that the client's rules for when the header is added are complicated, and it can also transparently add other headers. So in order to _accurately_ show the real headers sent to the server, we need to use the `httptrace` package.

There are a handful of other small changes in here, because I saw this errant content-length issue when examining the output of tests run against a grpc-java server. The conformance server implementation was incorrectly leaving `Host` empty in the `ServerCompatResponse`, and yet everything worked (except that the trace printed the URL without a host name). I've changed it to explicitly use "127.0.0.1" as a default if the server does not provide a host. And I also added some "TE: Trailers" request headers to the gRPC raw requests in test cases, because grpc-java servers complain in the logs if the requests do not include this. (All proper gRPC clients _do_ include it.)

### DIFF
--- a/internal/app/connectconformance/server_runner.go
+++ b/internal/app/connectconformance/server_runner.go
@@ -41,6 +41,8 @@ import (
 //
 // If isReferenceServer is true, then the server's stderr will be examined as well, to
 // record out-of-band feedback about the client requests.
+//
+//nolint:gocyclo
 func runTestCasesForServer(
 	ctx context.Context,
 	isReferenceClient bool,
--- a/internal/app/connectconformance/server_runner.go
+++ b/internal/app/connectconformance/server_runner.go
@@ -151,7 +151,7 @@ func runTestCasesForServer(
 		req := proto.Clone(testCase.Request).(*conformancev1.ClientCompatRequest) //nolint:errcheck,forcetypeassert
 		req.Host = resp.Host
 		if req.Host == "" {
-			req.Host = "127.0.0.1"
+			req.Host = internal.DefaultHost
 		}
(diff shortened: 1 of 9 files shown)


internal/app/connectconformance/server_runner.go
```diff
@@ -148,6 +150,9 @@ func runTestCasesForServer(
 		}
 		req := proto.Clone(testCase.Request).(*conformancev1.ClientCompatRequest) //nolint:errcheck,forcetypeassert
 		req.Host = resp.Host
+		if req.Host == "" {
+			req.Host = "127.0.0.1"
```
smaye81, line 154: We actually have this as a `const` defined in `internal/config.go`.

internal/app/referenceserver/server.go
```diff
@@ -102,6 +102,9 @@ func run(ctx context.Context, referenceMode bool, args []string, inReader io.Rea
 	if err != nil {
 		return err
 	}
+	if actualHost == "" || actualHost == "0.0.0.0" {
+		actualHost = "127.0.0.1"
```
smaye81, line 106: Ditto here re: `const`

smaye81 (approved):

PR #823, opened 2024-03-11
Add tracing support to gRPC impls by sniffing HTTP/2 frames on underlying TCP socket

### LINKED ISSUE #817: No traces generated from grpc client/server impls
The `--trace` flag has no effect for tests that are run using `(grpc client impl)` and `(grpc server impl)`.

jhump:
This adds a new tracing capability -- to wrap a `net.Conn` that represents an HTTP/2 connection and examine all of the HTTP/2 traffic to assemble a trace. This is used to instrument a gRPC client and server to support tracing, since they include their own HTTP/2 implementation instead of using `net/http`.

The only testing I've done is to set breakpoints in an IDE to make sure the trace events were flowing. I'll probably add some very basic tests of the tracing (both the new HTTP/2 stuff and the existing `net/http` stuff) and push them to this PR, just so we have more comfort that this stuff doesn't stays functional and doesn't regress since it is admittedly pretty complicated and not otherwise covered by tests.

This will resolve #817.

### DIFF
--- a/docs/configuring_and_running_tests.md
+++ b/docs/configuring_and_running_tests.md
@@ -405,10 +405,6 @@ protocol, it gets run against both reference implementations, as a way of furthe
 interoperability with the ecosystem. (Note that the standard reference implementations _also_ support
 the gRPC protocol; so the gRPC test cases are repeated with a different server.)
 
-Note that an HTTP trace, for failing tests when the `--test` flag is used, will only available with the
-primary reference implementations. Test cases that fail against the gRPC implementations will not have
-a trace (only the primary reference implementations are instrumented to capture such traces).
-
 ### Selecting Test Cases
 
 The `connectconformance` test runner supports four different options for selecting which test cases to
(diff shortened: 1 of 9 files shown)


smaye81 (approved):

jhump:
@smaye81, I just added some tests for the tracing stuff, and they caught a few things. Most of what they caught was non-deterministic ordering of events (which could even lead to events being dropped).

So this changes some of the locking in several places to make larger chunks of logic atomic -- for example the `handleFrame` method of the tracing `net.Conn` would acquire the lock, query a map, unlock, and then possibly have to acquire the lock again for a second step. Now everything is done atomically with a single lock acquisition. This was also true in the trace builder's `add` method -- it could do everything under lock and then, when it optionally called `build` from a defer function, it would re-acquire the lock. It's been restructured to only need a single lock acquisition now.

The new test is a bit of a beast, but it verifies all of the ways to instrument a client and server, both with middleware or with wrapping an HTTP/2 `net.Conn` socket.

smaye81:
lgmt, ship it!

PR #824, opened 2024-03-12
Fix `deflate` algorithm to adhere to RFC 1950

jchadwick-buf:
We don't default to testing `deflate` right now, as far as I can tell, but this is a pretty potent footgun so I think it is definitely worth getting right.

Though it is not explicitly specified anywhere I can find, [gRPC seems to follow HTTP in treating `deflate` as RFC 1950 (zlib) instead of RFC 1951 (deflate)](https://github.com/grpc/grpc/blob/f99b8b5bc4a89d37f86bf063999ec22e44312663/src/core/lib/compression/message_compress.cc#L131).

The `deflateInit2` function's documentation can be found online [here](https://www.zlib.net/manual.html). But briefly:

- If `windowBits` is [-15, -8], raw deflate (RFC 1951) is used.
- If `windowBits` is [8, 15], deflate with zlib header (RFC 1950) is used.
- If `windowBits` is [24, 31], deflate with gzip header (RFC 1952) is used.

Following this, we can see that gRPC will use either gzip (for the `gzip` encoding) or zlib (for the `deflate` encoding.)

### DIFF
--- a/internal/compression/deflate.go
+++ b/internal/compression/deflate.go
@@ -30,14 +30,24 @@ type deflateDecompressor struct {
 }
 
 func (c *deflateDecompressor) Read(bytes []byte) (int, error) {
+	if c.reader == nil {
+		return 0, io.EOF
+	}
 	return c.reader.Read(bytes)
 }
 func (c *deflateDecompressor) Reset(rdr io.Reader) error {
-	var err error
-	c.reader, err = zlib.NewReader(rdr)
-	return err
+	reader, err := zlib.NewReader(rdr)
+	if err != nil {
+		c.reader = &errorDecompressor{err: err}
+		return err
+	}
+	c.reader = reader
+	return nil
 }
 func (c *deflateDecompressor) Close() error {
+	if c.reader == nil {
+		return nil
+	}
 	return c.reader.Close()
 }
 
--- a/internal/compression/deflate.go
+++ b/internal/compression/deflate.go
@@ -15,14 +15,17 @@
 package compression
 
 import (
-	"compress/flate"
+	"compress/zlib"
 	"errors"
 	"io"
 
 	"connectrpc.com/connect"
 )
 
-// deflateDecompressor is a thin wrapper around a flate Reader.
+// deflateDecompressor is a thin wrapper around a zlib Reader. Note that due to
+// an unfortunate misnomer with the RFC 2616 specification, HTTP deflate is
+// actually RFC 1950 with zlib headers, rather than RFC 1951. gRPC uses the same
+// nomenclature.
 type deflateDecompressor struct {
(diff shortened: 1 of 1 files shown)


internal/compression/deflate.go
```diff
@@ -31,33 +33,21 @@ func (c *deflateDecompressor) Read(bytes []byte) (int, error) {
...
-		// This should never happen as the returned type from flate should always
-		// implement Resetter, but the check is here as a safeguard just in case.
-		// This error would be a very exceptional / unexpected occurrence.
-		return errors.New("deflate reader is not able to be used as a resetter")
-	}
-	// Mimics NewReader internal logic, which initializes the internal dict to nil
-	return resetter.Reset(rdr, nil)
+	var err error
+	c.reader, err = zlib.NewReader(rdr)
+	return err
```
jhump, line 38: This could leave `c.reader` as nil. Maybe it would be safest to do something like so:
```suggestion
	decomp, err := zlib.NewReader(rdr)
	if err != nil {
		c.reader = errDecompressor{err: err}
		return err
	}
	c.reader = decomp
	return nil
```

An alternative would be to put a nil check for `c.reader` in the other methods, to return reasonable errors instead of letting them panic.

internal/compression/deflate.go
```diff
@@ -31,33 +33,21 @@ func (c *deflateDecompressor) Read(bytes []byte) (int, error) {
...
 func (c *deflateDecompressor) Close() error {
 	return c.reader.Close()
 }
 
 // NewDeflateDecompressor returns a new deflate Decompressor.
 func NewDeflateDecompressor() connect.Decompressor {
-	return &deflateDecompressor{
-		reader: flate.NewReader(nil),
-	}
+	return &deflateDecompressor{}
```
jhump, line 46: Ditto -- could either initialize `reader` field to something like `errorDecompressor{err: io.EOF}` or could add appropriate nil guards in other methods, to avoid panics.

jhump (approved):
I think it would be nice to add minor safeguards against nil-dereference panics.

But otherwise, LGTM!

internal/compression/deflate.go
```diff
@@ -47,17 +50,17 @@ func (c *deflateDecompressor) Close() error {
 
 // NewDeflateDecompressor returns a new deflate Decompressor.
 func NewDeflateDecompressor() connect.Decompressor {
+	reader, err := zlib.NewReader(nil)
```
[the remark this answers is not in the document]
jchadwick-buf: Fixed.

internal/compression/deflate.go
jchadwick-buf: I added nil checks in the other methods, too. That should cover our bases pretty well I think.

PR #825, opened 2024-03-13
Use connect.ErrorWriter to write a better error when test case name header is missing

jhump:


smaye81 (approved):
Yeah this is perfect. Should help with the 400 issues.

PR #828, opened 2024-03-13
Test runner generates self-signed cert that servers-under-test can use

jhump:
This allows servers-under-test to just use a given PEM-encoded certificate, instead of having to generate their own.


jhump:
cc @jchadwick-buf, @srikrsna-buf

smaye81 (approved):
Should probably also document the change to `ServerCompatRequest` in https://github.com/connectrpc/conformance/blob/main/docs/testing_servers.md.

jhump:
> Should probably also document the change to `ServerCompatRequest` in https://github.com/connectrpc/conformance/blob/main/docs/testing_servers.md.

Done!

PR #826, opened 2024-03-13
Reconcile test cases w/ latest spec changes and fixes from latest connect-go

jhump:
This updates to the latest connect-go release and reconciles the test cases with the latest spec changes (from https://github.com/connectrpc/connectrpc.com/pull/130). It also enables some other tests that were waiting on connect-go fixes and cleans up the "known failing" lists based on these fixes and the new test cases. Finally, there are some tweaks to gRPC test cases, related to relaxing error code assertions.

This should be the last thing we need before releasing v1.0.0-rc4.

Resolves #810.

This also brings in [a fix in connect-go](https://github.com/connectrpc/connect-go/pull/709) that resolves #813.

PR #835, opened 2024-03-19
Add GitHub action workflow to automatically build/publish release artifacts after tag is created

jhump:
I tested this configuration with a personal repo to verify it does what we want. This allows us to hand-curate the release notes in the GitHub UI, but still have `goreleaser` create the artifacts that are part of the release.

This also adds a `RELEASE.md` doc to describe the release process, and it's similar to connect-go (which will likely soon have a similar "release" action to publish the `protoc-gen-connect-go` binary as a release artifact) and connect-kotlin (which uses a "release" action to publish to Maven central).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        # When changing this, don't forget to also update
+        # the version of Go used in the release.yaml workflow.
         go-version: [1.20.x, 1.21.x, 1.22.x]
     steps:
       - name: Checkout Code
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.20.x, 1.21.x]
+        go-version: [1.20.x, 1.21.x, 1.22.x]
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
@@ -41,10 +41,10 @@ jobs:
         # conflicting guidance, run only on the most recent supported version.
         # For the same reason, only check generated code on the most recent
         # supported version.
-        if: matrix.go-version == '1.21.x'
+        if: matrix.go-version == '1.22.x'
         run: make checkgenerate && make lint
       - name: Check Release
         # We'll only be building releases w/ latest Go, so we only need to
         # test that it can be built w/ latest.
-        if: matrix.go-version == '1.21.x'
+        if: matrix.go-version == '1.22.x'
         run: make checkrelease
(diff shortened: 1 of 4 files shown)


jhump:
Ping, @smaye81.

After #826 is merged, we should be ready for v1.0.0-rc4 (hopefully final release candidate). And I'd like to try out this change with that next release.

RELEASE.md
```diff
@@ -0,0 +1,49 @@
...
+
+Using the Github UI, [create a new release](https://github.com/connectrpc/conformance/releases/new)
+like so:
+* Under “Choose a tag”, type in “vX.Y.Z” to create a new tag for the release upon publish.
+  Note: The release job does infer the version from the release string and expects versions
+  to start with `v` (e.g. `v0.1.0`).
+* Target the main branch.
+* Title the Release the same as the tag: “vX.Y.Z”.
+* Click “Set as latest release”.
+* If this is a release candidate, or any other kind of pre-release, click "Set as a pre-release".
```
akshayjshah, line 17: Can we specifically mention that pre-1.0 releases are _not_ automatically prereleases? Across our OSS, we settled on treating `v0.9.5` as a regular release. I assume that also applies to RCs for the 1.0, since we _do_ want to feature those releases on the repository homepage?

.github/workflows/ci.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.20.x, 1.21.x]
+        go-version: [1.20.x, 1.21.x, 1.22.x]
```
akshayjshah, line 15: Can we leave a comment reminding maintainers to also keep `release.yaml` updated when changing Go versions?

akshayjshah (approved):

RELEASE.md
jhump: I didn't realize that RCs should not be marked as pre-releases. For Go libraries anyway, using `go get -u` won't get release candidates, so it seemed okay to align the GitHub release state the same way, by marking them as pre-releases.

I can wordsmith a little on this.

.github/workflows/ci.yaml
jhump: Great idea

jhump:
@akshayjshah, I was just looking at this and realized I totally forgot to describe the process to update the embedded version number via PR. So I just pushed a commit with more words, mostly copied from similar language for connect-go releases.

RELEASE.md
jhump: I added a single statement about pre-v1.0.0 releases. But would like to confirm: do we want release candidates to _not_ be release candidates, too?

akshayjshah: As far as I can tell, the only effect of this checkbox for Go libraries and binaries is controlling whether or not they're featured on the repository homepage. It seems that we want to advertise the existence of v1.0-rc4, right? We might feel differently about RCs for v2, if we ever get there - we'd want most users to continue using 1.x releases instead of blindly updating to something unstable.

akshayjshah (approved):

PR #844, opened 2024-04-04
Improve test runner behavior when client times out, crashes, or exits prematurely

jhump:
While troubleshooting some other test cases, and trying to figure out which test cases were causing a client impl to hang, I noticed a issues with the output of the test runner when this sort of thing happens:

1. If `-v` or `--vv` is used, it can be seen that the runner will print that it's trying to run more test cases and send one or more tests to the client, even though it has failed (timed out usually, but can also happen if it crashes). So I've fixed it up so that once the client fails or times out, we don't try to do anything else and don't print anything else about trying to run other test cases.
2. If `--vv` is used, the pending test cases (at least one of which would be the culprit for the client hanging or crashing) all get printed in "Received response for ..." messages, even though no response was actually received from the client.
3. When the client timed out or crashed, the process would abort and not bother showing any other results. This can be kind of a pain -- it is often useful to see the results that _were_ computed for all of the tests that completed prior to the client hanging or crashing.
4. Finally, unrelated to clients hanging/crashing: the `-v` output that printed the number of server configurations was misleading. For example, when running `make runconformance` in this repo, it prints 19 configurations for the reference client and server. _But_ it then proceeds in printing "Running X tests with ..." 20 or 22 times. This is because it actually starts different processes when testing with the gRPC implementations. So there may be 19 distinct configurations like `{HTTP_VERSION_2, PROTOCOL_GRPC, TLS:false}`, there actually 20 or 22 servers started, some being used with the gRPC reference implementation.

All of these issues are resolved in this PR.

### DIFF
--- a/internal/app/connectconformance/client_runner.go
+++ b/internal/app/connectconformance/client_runner.go
@@ -195,7 +195,7 @@ func (c *clientProcessRunner) consumeOutput() {
 			if err == nil || errors.Is(err, io.EOF) {
 				err = errNoOutcome
 			}
-			action(key, nil, &errFailedToGetResult{err})
+			action(key, nil, &failedToGetResultError{err})
 			delete(c.pendingOps, key)
 		}
 	}()
@@ -229,14 +229,14 @@ func (c *clientProcessRunner) consumeOutput() {
 	}
 }
 
-type errFailedToGetResult struct {
+type failedToGetResultError struct {
 	err error
 }
 
-func (e *errFailedToGetResult) Error() string {
+func (e *failedToGetResultError) Error() string {
 	return fmt.Sprintf("failed to get result from client: %v", e.err)
 }
 
-func (e *errFailedToGetResult) Unwrap() error {
+func (e *failedToGetResultError) Unwrap() error {
 	return e.err
 }
--- a/internal/app/connectconformance/client_runner.go
+++ b/internal/app/connectconformance/client_runner.go
@@ -32,8 +32,11 @@ const (
 	maxClientResponseSize = 16 * 1024 * 1024 // 16 MB
 )
 
-var errClosed = errors.New("send-to-client is closed")
-var errDuplicate = errors.New("duplicate test case")
+var (
+	errClosed    = errors.New("could not send request: client stdin is closed")
+	errDuplicate = errors.New("duplicate test case")
+	errNoOutcome = errors.New("no outcome ever received")
+)
 
 type clientRunner interface {
 	sendRequest(req *conformancev1.ClientCompatRequest, whenDone func(string, *conformancev1.ClientCompatResponse, error)) error
@@ -124,7 +127,7 @@ func (c *clientProcessRunner) sendRequest(req *conformancev1.ClientCompatRequest
 		}
 
 		if errors.Is(err, io.ErrClosedPipe) {
-			err = errors.New("could not write request: client closed stdin")
+			err = errClosed
 		}
 		c.err.CompareAndSwap(nil, &err)
 		return err
@@ -190,9 +193,9 @@ func (c *clientProcessRunner) consumeOutput() {
 		for key, action := range c.pendingOps {
 			err := reasonForReturn
 			if err == nil || errors.Is(err, io.EOF) {
-				err = fmt.Errorf("client never provided response for test case %q", key)
+				err = errNoOutcome
 			}
-			action(key, nil, err)
+			action(key, nil, &errFailedToGetResult{err})
 			delete(c.pendingOps, key)
 		}
 	}()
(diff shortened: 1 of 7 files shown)


internal/app/connectconformance/client_runner.go
```diff
@@ -32,8 +32,11 @@ const (
 	maxClientResponseSize = 16 * 1024 * 1024 // 16 MB
 )
 
-var errClosed = errors.New("send-to-client is closed")
-var errDuplicate = errors.New("duplicate test case")
+var (
+	errClosed    = errors.New("could not send request: client stdin is closed")
+	errDuplicate = errors.New("duplicate test case")
+	errNoOutcome = errors.New("no outcome ever received")
```
jhump, line 38: This is just for consistency with errors. We had two places trying to indicate the same thing, but using different error wording. So now I've factored it into an error var that is used from both places.

internal/app/connectconformance/client_runner.go
```diff
@@ -225,3 +228,15 @@ func (c *clientProcessRunner) consumeOutput() {
 		action(resp.TestName, resp, nil)
 	}
 }
+
+type errFailedToGetResult struct {
```
jhump, line 232: This marker type is used to avoid printing "Receive response for ..." verbose messages. When the test runner sees this as the cause of failure for a test case, it knows that it didn't _actually_ receive any response.

internal/app/connectconformance/connectconformance.go
```diff
@@ -115,10 +115,13 @@ func Run(flags *Flags, logPrinter internal.Printer, errPrinter internal.Printer)
...
 	results, err := run(configCases, knownFailing, knownFlaky, runPatterns, skipPatterns, allSuites, logPrinter, errPrinter, flags)
-	if err != nil {
+	if results == nil {
 		return false, err
 	}
-	return results.report(logPrinter), nil
+	if err != nil {
+		errPrinter.Printf("%v", err)
+	}
+	return results.report(logPrinter) && err == nil, nil
```
jhump, line 124: This is so that we can print the results, even after the client times out or crashes. We only return an error to the caller when there's an error _and_ no results. If we have both, report the error and the results. The bool value that is returned indicates the process should exit non-zero. So now we have to tell the caller to exit zero only when the results indicate no failures _and_ the error is nil.

internal/app/connectconformance/connectconformance.go
```diff
@@ -204,27 +207,39 @@ func run( //nolint:gocyclo
 	}
 
 	filter := newFilter(run, skip)
+	var filteredTestCount int
+	type serverConfig struct {
+		serverInstance
+		isGrpcClient, isGrpcServer bool
+	}
```
jhump, line 214: This is for two things:
1. It fixes the reporting of the number of server configurations, by including the gRPC implementations as separate configurations.
2. We now need to know the total number of test cases (after filtering), even when we're not using verbose printing. This is so that the results know if there were any test cases skipped or not run (due to client hanging/crashing).

internal/app/connectconformance/connectconformance.go
```diff
@@ -356,6 +371,15 @@ func run( //nolint:gocyclo
 						return err
 					}
 
+					// Double-check that client is still running before spawning a server process.
```
jhump, line 374: Oops, we were previously printing the "Running X tests with ..." message, even if we didn't actually run anything because the client stopped. A simple matter of putting that verbose logging _after_ this check, instead of doing it before.

internal/app/connectconformance/server_runner.go
```diff
@@ -236,8 +237,9 @@ func runTestCasesForServer(
 			wg.Done() // call it explicitly since callback above won't be invoked
 			// client pipe broken: mark remaining tests, including this one, as failed
 			for j := i; j < len(testCases); j++ {
-				results.setOutcome(testCases[j].Request.TestName, true, err)
+				results.setOutcome(testCases[j].Request.TestName, true, &errCouldNotRun{err})
 			}
+			break
```
jhump, line 242: This is the reason that, even after the client stopped, the `--vv` output would continue to show lots of "Sending request for ..." messages, because after detecting the error we didn't actually break out of this loop. Oops.

internal/app/connectconformance/server_runner.go
```diff
@@ -251,5 +253,17 @@ func runTestCasesForServer(
 	}
 
 	// If there are any tests without outcomes, mark them now.
-	results.failRemaining(testCases, errors.New("no outcome received from the client"))
+	results.failRemaining(testCases, &errFailedToGetResult{errNoOutcome})
+}
+
+type errCouldNotRun struct {
```
jhump, line 259: This marker type is used to account for tests that were never sent to the client when reporting that in the results. That figure includes all test cases whose outcome is this kind of error as well as all test cases that have no result at all.

smaye81 (approved):

PR #847, opened 2024-04-05
Must percent-encode gRPC message

jhump:
I thought I had tested this change against a bunch of other implementations, but I clearly missed this. But in #840, the code was not correctly percent-encoding the "grpc-message" trailer. It turns out that most implementations (including grpc-java, grpc-go, and connect-go) were all lenient in the face of this issue, which is _how_ I missed this. Looks like we'll need a 1.0.1 😒 

I think what might have been happening is that `net/http` was _dropping_ the invalid header (since one of the tests has a newline in the message, which is not allowed in headers) but it was still being picked up from the `grpc-status-details-bin` header. But it caused an error in connect-kotlin. (I could have _sworn_ I tested that commit against kotlin last week, but I guess I didn't actually do that correctly.)

So... this fixes the issue: when the reference server synthesizes a raw response, to reply with independent header and trailer metadata, it now correctly encodes the "grpc-message" key.

The _bulk_ of this PR is actually adding new checks in the reference client that could have caught this issue: it examines the "grpc-status", "grpc-message", and "grpc-status-details-bin" to make numerous checks to verify that the server implementation adheres to the specs. It also applies similar checks to other binary metadata.

There are a handful of small other changes in here too that I addresses while troubleshooting and fix this.

### DIFF
--- a/internal/app/connectconformance/results.go
+++ b/internal/app/connectconformance/results.go
@@ -174,14 +174,18 @@ func (r *testResults) assert(
 		// "error metadata". The conformance client should record those as trailers when
 		// sending back a ClientResponseResult message.
 
-		// So first we see if normal attribute succeeds
+		// So first we see if normal attribution succeeds.
 		metadataErrs := checkHeaders("response headers", expected.ResponseHeaders, actual.ResponseHeaders)
 		metadataErrs = append(metadataErrs, checkHeaders("response trailers", expected.ResponseTrailers, actual.ResponseTrailers)...)
 		if len(metadataErrs) > 0 {
-			// That did not work. So we test to see if client attributed them all as trailers.
+			// That did not work. So we test to see if client attributed them all as headers
+			// or all as trailers.
 			merged := mergeHeaders(expected.ResponseHeaders, expected.ResponseTrailers)
-			if allTrailersErrs := checkHeaders("response metadata", merged, actual.ResponseTrailers); len(allTrailersErrs) != 0 {
-				// That check failed also. So the received headers/trailers are incorrect.
+			allHeadersErrs := checkHeaders("response metadata", merged, actual.ResponseHeaders)
+			allTrailersErrs := checkHeaders("response metadata", merged, actual.ResponseTrailers)
+			// That check failed also. So see if client attributed them all as headers.
+			if len(allHeadersErrs) != 0 && len(allTrailersErrs) != 0 {
+				// These checks failed also. So the received headers/trailers are incorrect.
 				// Report the original errors computed above.
 				errs = append(errs, metadataErrs...)
 			}
--- a/internal/app/connectconformance/testsuites/data/grpc_web_client_trailers.yaml
+++ b/internal/app/connectconformance/testsuites/data/grpc_web_client_trailers.yaml
@@ -122,8 +122,6 @@ testCases:
                   value: [ "*" ]
                 - name: content-type
                   value: [ "application/grpc-web" ]
-                - name: x-custom-header
-                  value: [ "foo" ]
                 - name: x-custom-trailer
                   value: [ "bing" ]
                 - name: grpc-status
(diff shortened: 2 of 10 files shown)


internal/app/connectconformance/testsuites/data/grpc_web_client_trailers.yaml
```diff
@@ -122,18 +122,13 @@ testCases:
...
                 - name: x-custom-trailer
                   value: [ "bing" ]
                 - name: grpc-status
                   value: [ "9" ]
                 - name: grpc-message
                   value: [ "error" ]
     expectedResponse:
-      responseHeaders:
-        - name: x-custom-header
-          value: [ "foo" ]
```
jhump, line 136: 🤦 this was just wrong. All of the metadata were in the same bucket in the trailers-only response, so the expected response should not have them split up. I just removed "headers" entirely from these since they don't make sense.

internal/app/referenceclient/impl.go
```diff
@@ -117,54 +117,10 @@ func (i *invoker) unary(
...
-		ResponseTrailers: trailers,
-		Payloads:         payloads,
-		Error:            protoErr,
-		HttpStatusCode:   statusCode,
-		Feedback:         feedback,
-	}, nil
+	return doUnary(ctx, req, i, i.client.Unary,
+		func(resp *conformancev1.UnaryResponse) *conformancev1.ConformancePayload {
+			return resp.Payload
+		})
```
jhump, line 123: I was making some small changes in here and realized that not only was this logic duplicated, but it was also subtly different between the two already. So I decided to address the TODO below to consolidate the logic.

internal/app/connectconformance/results.go
```diff
@@ -174,14 +174,18 @@ func (r *testResults) assert(
...
+			// That did not work. So we test to see if client attributed them all as headers
+			// or all as trailers.
 			merged := mergeHeaders(expected.ResponseHeaders, expected.ResponseTrailers)
-			if allTrailersErrs := checkHeaders("response metadata", merged, actual.ResponseTrailers); len(allTrailersErrs) != 0 {
-				// That check failed also. So the received headers/trailers are incorrect.
+			allHeadersErrs := checkHeaders("response metadata", merged, actual.ResponseHeaders)
+			allTrailersErrs := checkHeaders("response metadata", merged, actual.ResponseTrailers)
+			// That check failed also. So see if client attributed them all as headers.
+			if len(allHeadersErrs) != 0 && len(allTrailersErrs) != 0 {
+				// These checks failed also. So the received headers/trailers are incorrect.
```
jhump, line 188: I changed this to be more strict in #805. But now I think that was a mistake, so am making this a little more lenient again, reverting it to what it looked like before.

internal/grpcutil/metadata.go
```diff
@@ -81,3 +81,39 @@ func AppendToOutgoingContext(ctx context.Context, src []*conformancev1.Header) c
 	}
 	return ctx
 }
+
+// PercentEncodeMessage percent-encodes the given string per the rules in the
+// gRPC spec for the "grpc-message" trailer value.
+func PercentEncodeMessage(msg string) string {
```
jhump, line 87: This was basically lifted from connect-go

internal/tracer/builder.go
```diff
@@ -116,7 +116,7 @@ func (b *builder) add(event Event) {
 		event.MessageIndex = b.reqCount
 		b.reqCount++
 	case *RequestBodyEnd:
-		if b.trace.Err != nil {
+		if b.trace.Err == nil {
```
jhump, line 119: Oof, these error conditions were just straight-up wrong 🤦. The intent was to set the error if it's not already set. But the inverted condition meant it was _never_ being set.

smaye81 (approved):

PR #855, opened 2024-04-08
Fix reference server issues related to CORS

jhump:
The reference server was already configured to use CORS and to be permissive. But there were some quirks to its config that could be improved (especially if anyone were to test using TLS client certs, which means we have to allow credentials in the cross-origin requests).

Also, the "raw response" capability of the reference server was interfering with the CORS middleware. So this fixes that, which in turn activates some alternate behavior in the gRPC Web JS client: apparently, some of the behavior observed was actually the result of failed preflight requests and not the intended behavior of the library 🤦.

On a positive note: it is no longer necessary/useful to add access control headers to raw responses. The CORS middleware now handles that correctly.

### DIFF
--- a/testing/grpcwebclient-known-failing.txt
+++ b/testing/grpcwebclient-known-failing.txt
@@ -30,9 +30,11 @@ gRPC-Web Unexpected Responses/**/ok-but-no-response
 # grpc-status header/trailer key.
 HTTP to RPC Code Mapping/**/bad-gateway
 HTTP to RPC Code Mapping/**/bad-request
-HTTP to RPC Code Mapping/**/forbidden
 HTTP to RPC Code Mapping/**/gateway-timeout
 HTTP to RPC Code Mapping/**/not-found
-HTTP to RPC Code Mapping/**/service-unavailable
 HTTP to RPC Code Mapping/**/too-many-requests
-HTTP to RPC Code Mapping/**/unauthorized
+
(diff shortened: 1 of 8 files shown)


testing/grpcwebclient-known-failing.txt
```diff
@@ -30,9 +30,11 @@ gRPC-Web Unexpected Responses/**/ok-but-no-response
...
 HTTP to RPC Code Mapping/**/gateway-timeout
 HTTP to RPC Code Mapping/**/not-found
-HTTP to RPC Code Mapping/**/service-unavailable
 HTTP to RPC Code Mapping/**/too-many-requests
-HTTP to RPC Code Mapping/**/unauthorized
+
+# The above table maps 409 and 412 to "unknown" RPC code. But this client
+# instead maps them to "aborted" and "failed_precondition" respectively.
+HTTP to RPC Code Mapping/**/conflict
+HTTP to RPC Code Mapping/**/precondition-failed
```
jhump, line 40: Apparently, these were all "unknown" before because of failed access-control checks 😢 

Fixing the CORS issues reveals three of the cases actually pass, but now there are two others that fail due to use of non-standard mappings. (By "non-standard", I mean mapping that aren't specified [here](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md).)

jhump:
cc: @timostamm, @srikrsna-buf

smaye81 (approved):

PR #856, opened 2024-04-08
When reference server is in HTTP 1.1 mode, don't allow negotiating HTTP/2 via TLS

### LINKED ISSUE #854: Reference server should not advertise HTTP/2 when running in HTTP/1.1
Reference server should not advertise `HTTP/2` when running in `HTTP/1.1`. In a browser environment, if the server support H2 and is secure, the browser will use H2 and there is no way to control this. If the reference server for H1 tests only supported H1, then the browser will only end up using H1.

jhump:
Some clients may not have the direct ability to _disable_ negotiating HTTP/2 if the server advertises its support during ALPN step of TLS handshake. So to prevent such clients from using HTTP/2, if HTTP 1.1. is the intent, we disable HTTP/2 in the reference server. (Sadly, we can't also disable HTTP 1.1 when HTTP/2 is the intent since Go's `net/http` doesn't provide a way to do that 🤷.)

This should resolve #854

### DIFF
--- a/internal/app/referenceserver/server.go
+++ b/internal/app/referenceserver/server.go
@@ -314,6 +314,8 @@ func newH1Server(handler http.Handler, listenAddr string, tlsConf *tls.Config) (
 		TLSConfig:         tlsConf,
 		ReadHeaderTimeout: 5 * time.Second,
 		ErrorLog:          nopLogger(),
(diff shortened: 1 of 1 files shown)


jhump:
cc @srikrsna-buf

smaye81 (approved):

PR #866, opened 2024-04-29
Update Go to 1.21 and quic-go to 0.43.0 (which now requires Go 1.21)

jhump:
Revisiting #759, I think it's okay to update Go in this repo. It's not a library, so we shouldn't have to worry about any users trying to import this module from a program using an older version of Go. And, since it's not a library, we can have a more strict policy on what versions of Go we support. The binary we include in releases is currently always built from latest (using go1.22.x image in CI), so there are no clear downsides here (other than needing to tweak required checks in branch configs, too).


internal/app/referenceclient/client.go
```diff
@@ -227,7 +227,7 @@ func invoke(ctx context.Context, req *conformancev1.ClientCompatRequest, referen
 		transport = &contextFixTransport{http3.RoundTripper{
 			DisableCompression: true,
 			TLSClientConfig:    tlsConf,
-			QuicConfig:         &quic.Config{MaxIdleTimeout: 20 * time.Second, KeepAlivePeriod: 5 * time.Second},
+			QUICConfig:         &quic.Config{MaxIdleTimeout: 20 * time.Second, KeepAlivePeriod: 5 * time.Second},
```
jhump, line 230: This is documented as one of the breaking changes in v0.43.0: https://github.com/quic-go/quic-go/releases/tag/v0.43.0

PR #874, opened 2024-05-23
Update test cases for consistency, backfill missing cancel tests

jhump:
After noticing some naming inconsistencies, I added a test that verifies that the embedded test files all conform to the naming conventions described in [the docs](https://github.com/connectrpc/conformance/blob/main/docs/authoring_test_cases.md#naming-conventions).

That found a few issues, so this PR fixes them all.

I noticed and changed a few other things while doing this other clean-up in the tests:
1. We were being lenient with gRPC implementations regarding an unknown content-type. That's because I thought this behavior (using a 415 HTTP status code) wasn't fully specified. But it turns out that [it is](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#:~:text=If%20Content%2DType%20does%20not%20begin%20with%20%22application/grpc%22%2C%20gRPC%20servers%20SHOULD%20respond%20with%20HTTP%20status%20of%20415%20(Unsupported%20Media%20Type).%20This%20will%20prevent%20other%20HTTP/2%20clients%20from%20interpreting%20a%20gRPC%20error%20response%2C%20which%20uses%20status%20200%20(OK)%2C%20as%20successful.)! So I've moved these all to a test suite that is not protocol-specific, since all three protocols now expect the same behavior.
2. I noticed a test that was commented out with a TODO. The change the TODO was contingent was on was already done, so the test could be re-enabled.
3. The new naming convention test revealed a naming issue in the cancellation tests. When I went to fix that, I also noticed there were some redundant cases in here: the `cancel-after-zero-responses` test cases are not actually different than the `cancel-after-close-send`. While the contents of the test case are slightly different, the behavior they induce in the client is the same, which is to cancel the RPC right after the request stream is closed. It seemed odd that we'd have a test for cancelling "after" receiving zero responses. That's really like a cancel before the first response is received, and that felt awkward to implement since ideally you simply check that number _after_ receiving a response (per the name of the field). So I opted to remove those and keep the `cancel-after-close-send` version.
3. The server-stream cancel tests only had the `cancel-after-zero-responses`, and no `cancel-after-close-send` version. So I "fixed" that -- to use `cancel-after-send` for consistency with the others -- only to realize that the reference clients did not actually implement it. That's when I also realized that there was no test for cancelling unary RPCs. (The reference clients also did not implement cancellation in the unary RPC flow.)

The biggest changes in here were to get all of the client implementations to handle the `cancel.after_close_send_ms` field in the input.


testing/grpcwebclient/browserscript.ts
```diff
@@ -53,6 +46,17 @@ async function runTestCase(data: number[]): Promise<number[]> {
   return Array.from(result.serializeBinary());
 }
 
+function log(msg: string) {
+  let now = new Date();
+  // eek, there must be a better way...
```
smaye81, line 51: Try:

`let timestamp = now.toLocaleTimeString("en-US", { hour12: false })`

smaye81 (approved):

testing/grpcwebclient/browserscript.ts
jhump: That doesn't include millisecond precision, which is what I was going for.

This function is actually unused currently, but I had added temporary log lines to troubleshoot something, and I needed millisecond precision to really understand the timing of some of the event callbacks.

internal/app/grpcclient/impl.go
```diff
@@ -400,19 +377,74 @@ func (i *invoker) unimplemented(
...
 
 	// Add the specified request headers to the request
 	ctx = grpcutil.AppendToOutgoingContext(ctx, ccr.RequestHeaders)
 
+	var protoErr *conformancev1.Error
+	payloads := make([]*conformancev1.ConformancePayload, 0, 1)
+
+	var headerMD, trailerMD metadata.MD
+
+	if timing.AfterCloseSendMs >= 0 {
```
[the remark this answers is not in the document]
jhump: I just pushed a commit that beefs up the doc comments for this, to hopefully make it clear. The recently-ish-added docs for testing clients already had some of this info. In that commit, I also reconciled the implementations in referenceclient and grpcclient with the pseudo-code in those docs (and found and fixed one omission in those docs). There are some other minor changes in the doc comments, just to get to consistency on whether to use hyphens or not, like "server-stream" vs. "server stream" (I chose the latter and updated everything else in the file to follow suit).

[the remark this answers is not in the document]
smaye81: Yep, looks great. Thanks for updating.

testing/grpcwebclient/browserscript.ts
jhump: I decided to just remove the function since it's not actually used right now. (Seemed like too much work to add more log lines and iterate on it until it looked good with a new formulation... We can apply something like this if/when we need it in the future.)

PR #877, opened 2024-05-29
Require that client implementations properly enforce timeouts

jhump:
Because the connect-go server framework will try to curtail work beyond the deadline, the timeout test cases weren't working quite as expected. They worked for unary, but for the streaming endpoints, the deadline would elapse and cause the server handler's stream interactions to return a `deadline_exceeded` error, which it would then return to the client. So the client was correctly reporting a `deadline_exceeded` error for the streaming tests, even when it didn't actually enforce a deadline.

So this change causes the reference server to _remove_ the timeout headers, so that the connect-go framework it is built on will not bother trying to enforce any client-provided deadline. In order to actually validate the client deadline in these cases, the timeout is parsed (by the same code that removes it from the headers) and is stored off separately (in a context value), for retrieval from the server handler.

cc @rebello95

### DIFF
--- a/internal/app/referenceserver/checks.go
+++ b/internal/app/referenceserver/checks.go
@@ -15,13 +15,16 @@
 package referenceserver
 
 import (
+	"context"
 	"errors"
 	"io"
+	"math"
 	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"connectrpc.com/conformance/internal"
 	"connectrpc.com/conformance/internal/compression"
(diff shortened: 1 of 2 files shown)


internal/app/referenceserver/checks.go
```diff
@@ -39,10 +42,29 @@ const (
...
+}
+
+type feedbackPrinter struct {
+	p            internal.Printer
+	testCaseName string
+}
+
+func (p *feedbackPrinter) Printf(format string, args ...any) {
+	p.p.PrefixPrintf(p.testCaseName, format, args...)
+}
```
jhump, line 64: I moved these up to conform to our style guide, which wants types (and their methods) to appear above functions.

PR #878, opened 2024-05-29
Use strict codecs in reference client and server

jhump:
This should make it easier to troubleshoot issues caused by accidentally using a non-conformant JSON format. For example, it turns out that Dart's [default JSON format](https://github.com/google/protobuf.dart/issues/220) is _not_ the format defined by the [proto3 mapping spec](https://protobuf.dev/programming-guides/proto3/#json) but some bespoke format that uses tag numbers instead of names.

Previously, a server or client would accept such messages and just ignore unrecognized fields (which could be all of them). This could result in hard-to-troubleshoot errors, since the server would behave like it got a partial or empty request message.

This will cause the codec to complain about the unrecognized field name, which should produce more intuitive errors.

We do similar for the Protobuf binary format, too, disallowing unrecognized fields in the data.

### DIFF
--- a/internal/app/referenceclient/client.go
+++ b/internal/app/referenceclient/client.go
@@ -261,7 +261,7 @@ func invoke(ctx context.Context, req *conformancev1.ClientCompatRequest, referen
 	case conformancev1.Codec_CODEC_PROTO:
 		// this is the default, no option needed
 	case conformancev1.Codec_CODEC_JSON:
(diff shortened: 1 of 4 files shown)


smaye81 (approved):

jhump:
@smaye81, I realized that it would likely be advantageous to do the same for the protobuf codec. Please take another look at the last commits.

PR #883, opened 2024-06-14
Update the Connect GET test suite configuration so it's only run when a client/server supports GET

### LINKED ISSUE #882: I don't think `supports_connect_get: false` works
While working on the Python implementation, I had this set in my config until I implemented:

See: https://github.com/mattrobenolt/connect-python/blob/ref/conformance/config.yaml#L16

But it's running tests such as `Connect with GET/HTTPVersion:1/Codec:CODEC_PROTO/TLS:false/success` which, I'd expect to not be attempted.

jhump:
The test suite was missing a configuration attribute that would cause it to be omitted if the client or server under test was configured as not supporting Connect GET.

Resolves #882.

### DIFF
--- a/internal/app/connectconformance/testsuites/data/connect_with_get.yaml
+++ b/internal/app/connectconformance/testsuites/data/connect_with_get.yaml
@@ -2,14 +2,15 @@ name: Connect with GET
 relevantProtocols:
   - PROTOCOL_CONNECT
 relevantCompressions:
-  # TODO - Clients only use compression for GET requests if the request is
-  # larger than a particular size. So since we aren't sending a request here
-  # that exceeds that, our requests won't use compression even if we specify
-  # it. So, we're limiting this test to identity
+  # TODO: The reference client only uses compression for GET requests if the
+  # request is larger than a particular size. So since we aren't sending a
+  # request here that exceeds that, our requests won't use compression even
+  # if we specify it. So, we're limiting this test to identity.
   - COMPRESSION_IDENTITY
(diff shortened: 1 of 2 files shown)


smaye81 (approved):

internal/app/connectconformance/testsuites/data/connect_with_get.yaml
```diff
@@ -2,14 +2,15 @@ name: Connect with GET
...
   - PROTOCOL_CONNECT
 relevantCompressions:
-  # TODO - Clients only use compression for GET requests if the request is
-  # larger than a particular size. So since we aren't sending a request here
-  # that exceeds that, our requests won't use compression even if we specify
-  # it. So, we're limiting this test to identity
+  # TODO: The reference client only uses compression for GET requests if the
+  # request is larger than a particular size. So since we aren't sending a
+  # request here that exceeds that, our requests won't use compression even
+  # if we specify it. So, we're limiting this test to identity.
```
jhump, line 8: Sorry for my compulsion to tweak stuff 😂

I don't actually know how other clients behave, but connect-kotlin behaves differently. So I thought it would be useful to clarify _which_ client behaves this way.

The fact that implementations are not consistent on how to enable compression in GET requests does make it hard to test this case. FWIW, I think the reason a client might behave like this is because the compression output may not be canonical/deterministic, which means things like caching by URL don't work well. So they only compress when necessary to fit the request into a "reasonably-sized" URL.

PR #904, opened 2024-08-02
Adds gRPC tests for trailers-only

jhump:
This mirrors the gRPC-Web test cases that also handle different forms of trailers in responses.

I didn't have to actually touch the reference server for this to work because "net/http" handlers actually do the right thing: if the handler doesn't write anything to the body (and also never calls `responseWriter.Flush()`), then the headers aren't actually sent until after the handler exits, and the HTTP/2 frame sequence will just be a HEADER frame with the END-STREAM flag set.


jhump:
cc @timostamm

smaye81 (approved):

PR #918, opened 2024-09-04
Add conformance test cases for handling empty (zero byte) messages

jhump:
Adds new test cases to make sure that servers correctly handle requests of zero bytes and clients handle responses of zero bytes. The responses required the use of "raw responses" in the test case, because conformance server implementations by default never send an empty response message, since they echo back the request information in the response.

This also does some tweaking of the test cases in `basic.yaml` to remove redundant/unnecessary test cases. I also tried adding "no-requests" test cases, to go along with existing "no responses" test cases. However, this uncovered an issue between the reference client and the grpc server that is still eluding me. So I've left those new test cases commented out for now and can uncomment this after I have time to debug them a little more.

Finally, this updates all of the client implementations to add an empty `ConformancePayload` message for a response, even if that response did not contain a `payload` field. This is specifically for handling these empty responses. This also was worth an update to the docs so we provide guidance to users that are implementing clients on how to handle these.


cmd/connectconformance/main.go
```diff
@@ -219,8 +219,6 @@ func run(flags *flags, cobraFlags *pflag.FlagSet, command []string) { //nolint:g
 			fatal(`Server command (after the "----") is empty.`)
 		}
 	default:
-		// TODO: support mode "both", which would allow the caller to supply both the
```
jhump, line 222: Oops, the TODO was already addressed, but just forgot to remove it in earlier change.

internal/app/connectconformance/connectconformance.go
```diff
@@ -281,22 +281,26 @@ func run( //nolint:gocyclo
...
-					return referenceclient.RunInReferenceMode(ctx, args, inReader, outWriter, errWriter, trace)
-				}),
+				start: runInProcess(
+					[]string{
+						"reference-client",
+						"-p", strconv.Itoa(int(flags.Parallelism)),
+					}, func(ctx context.Context, args []string, inReader io.ReadCloser, outWriter, errWriter io.WriteCloser) error {
+						return referenceclient.RunInReferenceMode(ctx, args, inReader, outWriter, errWriter, trace)
+					},
+				),
```
jhump, line 291: Sorry if this is too noisy. I tweaked the formatting of this block and similar ones below which made it easier for me to read.

internal/app/connectconformance/connectconformance_test.go
```diff
@@ -100,8 +100,8 @@ func TestRun(t *testing.T) {
 	allPermutations := testCaseLib.allPermutations(true, true)
 	expectedNumCases := len(allPermutations)
 
-	// 208 test cases as of this writing, but we will likely add more
-	require.GreaterOrEqual(t, expectedNumCases, 208)
+	// 194 test cases as of this writing, but we will likely add more
+	require.GreaterOrEqual(t, expectedNumCases, 194)
```
jhump, line 104: I removed some redundant test cases from `basic.yaml`, which resulted in this number falling.

internal/app/connectconformance/test_case_library.go
```diff
@@ -559,16 +559,6 @@ func populateExpectedResponse(testCase *conformancev1.TestCase) error {
...
 
-	// TODO - This is just a temporary constraint to protect against panics for now.
-	// Eventually, we want to be able to test client and bidi streams where there are no request messages.
-	// The potential plan is for server impls to produce (and the code below to expect) a single response
-	// message in this situation, where the response data value is some fixed string (such as "no response definition")
-	// and whose request info will still be present, but we expect it to indicate zero request messages.
-	if len(testCase.Request.RequestMessages) == 0 {
-		testCase.ExpectedResponse = &conformancev1.ClientResponseResult{}
-		return nil
-	}
```
jhump, line 570: We now treat zero request messages as effectively the same as a request that has no response definition field. I did this so that I could add "no-requests" test cases, to test client and bidi streams where the request stream is empty. (However, I had to leave those new test cases commented out due to issues I ran into and haven't had a chance to finish debugging...)

internal/app/connectconformance/testsuites/data/basic.yaml
```diff
@@ -19,17 +19,6 @@ testCases:
...
-    testName: unary/empty-definition
-    streamType: STREAM_TYPE_UNARY
-    requestMessages:
-    - "@type": type.googleapis.com/connectrpc.conformance.v1.UnaryRequest
-      responseDefinition:
-- request:
-    testName: unary/no-definition
-    streamType: STREAM_TYPE_UNARY
-    requestMessages:
-    - "@type": type.googleapis.com/connectrpc.conformance.v1.UnaryRequest
```
jhump, line 32: These almost felt more like test cases for the conformance implementation itself and not actually exercising anything interesting as far as the RPC interactions they describe and functionality in other implementations that we are trying to verify.

For one, we treat a missing response definition the same as an empty one, so these two are redundant with one another. But for unary (and client-stream below), where the server _must_ always furnish a response, the scenario these test cases present didn't seem particularly useful.

internal/app/connectconformance/testsuites/data/basic.yaml
```diff
@@ -80,22 +69,20 @@ testCases:
...
-    - "@type": type.googleapis.com/connectrpc.conformance.v1.ServerStreamRequest
-      responseDefinition:
-        responseHeaders:
-        - name: x-custom-header
-          value: ["foo"]
-        responseTrailers:
-        - name: x-custom-trailer
-          value: ["bing"]
-- request:
-    testName: server-stream/no-definition
```
jhump, line 95: In this case, these two were similar (just like the two unary cases above that were removed). But I left one of them because they do check something interesting for server and bidi streams: response streams that have zero messages.

internal/app/referenceclient/impl.go
```diff
@@ -184,14 +184,11 @@ func doUnary[ReqT, RespT any, Req pointerMessage[ReqT]](
 		trailers = internal.ConvertToProtoHeader(connectErr.Meta())
 		protoErr = internal.ConvertConnectToProtoError(connectErr)
 	} else {
-		// If the call was successful, get the headers and trailers
+		// If the call was successful, get the returned payloads
+		// and the headers and trailers
+		payloads = append(payloads, getPayload(resp.Msg))
```
jhump, line 189: This was minor cleanup. As it was, it would have return a slice of zero payloads, even though there was a single response message (so it should have been a slice of one).

Adding nil is okay here. In fact, the client-stream case looked exactly like this before and worked just fine. So I just updated this to look exactly like the code in the client-stream handler.

smaye81 (approved):

PR #921, opened 2024-09-10
Add tests for connect error detail debug data

timostamm:
Connect clients must not depend on data in the "debug" key when deserializing details from an Error or EndStreamResponse. 

This PR adds a test where the "debug" key contains data that does not match the error detail, and is not a JSON object (see clarification regarding the property type here [here](https://github.com/connectrpc/connectrpc.com/pull/199)).


jhump (approved):
Maybe worth also adding cases where the `debug` attribute is completely absent?

timostamm:
> Maybe worth also adding cases where the `debug` attribute is completely absent?

Good call. Added in https://github.com/connectrpc/conformance/pull/921/commits/f028979c0741463d08ab220d00f7c2cf226790d7.

internal/app/connectconformance/testsuites/data/connect_client_error_endstream.yaml
```diff
@@ -107,6 +107,73 @@ testCases:
       error:
         code: CODE_OUT_OF_RANGE
         message: oops
+  - request:
+      # Clients must not depend on data in the "debug" key when deserializing details.
+      testName: error/disregard-detail-debug
```
jhump, line 112: I think the test cases are typically named in a way that describe the scenario, vs describing the implementation's expected behavior. How about `invalid-details-debug` instead of `disregard-detail-debug`? (Sorry I didn't mention this with previous remark.)

timostamm: 👍  There are a couple of existing tests named "allow-unrecognized...". Doesn't seem worth renaming them, but renamed the new ones in https://github.com/connectrpc/conformance/pull/921/commits/7a9e8c2f105f89a11a8c84c04842f2f34ae32662.

jhump (approved):

PR #922, opened 2024-09-10
Add tests for omitted message in connect error

timostamm:
The "message" property in a Connect Error or EndStreamResponse is optional. This PR adds tests for clients to make sure they are able to parse a response where "message" is omitted.


internal/app/connectconformance/testsuites/data/connect_client_error_endstream.yaml
```diff
@@ -83,6 +83,25 @@ testCases:
       error:
         code: CODE_UNAVAILABLE
         message: oops
+  - request:
+      testName: error/message-omitted
```
jhump, line 87: For consistency with above test case (named `missing-code`), how about `missing-message` instead?

timostamm: Code is required, but message is optional, so it isn't really missing. Maybe that's splitting hairs... I've renamed to `missing-message` for consistency, and added a comment explaining it's optional.

PR #933, opened 2024-10-07
Update buf CI GitHub action

emcfarlane:
Update the GitHub action to use `buf-action`. This has improved support for handling syncing GitHub repositories with the BSR. See the [default behaviour](https://github.com/bufbuild/buf-action?tab=readme-ov-file#default-behavior). Changes include:
- On each push to the repository the event will be synced. This includes the git-metadata of the push.
- Formatting has been added as a check.


.github/workflows/buf-ci.yaml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          path: head
+      - uses: actions/checkout@v4
+        with:
+          path: base
+          ref: ${{ github.event.pull_request.base.sha }}
+      - uses: bufbuild/buf-action@v1
+        with:
+          input: head/proto
+          breaking_against: base/proto
```
emcfarlane, line 24:
```suggestion
          input: proto
          breaking_against: 'buf.build/connectrpc/conformance'
```
This would also work, without the path specified in the checkouts above. Reason to prefer using the base branch is for branches on non base branches.

.github/workflows/buf-ci.yaml
```diff
@@ -0,0 +1,25 @@
+name: Buf CI
```
emcfarlane, line 1: Move to `buf-ci.yaml` to avoid `buf.yaml`

.github/workflows/buf-ci.yaml
jhump: It _needs_ to be the BSR for correctness, because these run _after_ commit/push.

So if you only check against the base branch:
1. You inadvertently push an incompatible change (such as a semantic merge conflict: two branches are both compatible with a shared base but not with each other). This check _fails_.
2. You push some other commit later that adds a new field, not yet addressing the compat problem in previous run. Now the check _passes_ because it is only checking compatibility with what's in Git, instead of compatibility with what has actually been published to the BSR. It should fail until a commit is made to _fix_ the compatibility issue identified in step 1.

Since the BSR module is what we recommend people consume and use for code generation, that is the thing where compatibility matters. Previous commit of main (or base branch) isn't really strong enough.

.github/workflows/buf-ci.yaml
jhump: Most commits don't touch the protos, so the old config had a path filter. I see the new action recommends the following config for this case:
```
on:
  push:
    paths:
      - '**.proto'
      - '**/buf.yaml'
      - '**/buf.lock'
      - '**/buf.md'
      - '**/README.md'
      - '**/LICENSE'
```
We probably also want to add `.github/workflows/buf-ci.yaml` to that list.

.github/workflows/buf-ci.yaml
emcfarlane: Agree, fixed.

.github/workflows/buf-ci.yaml
emcfarlane: Added.

jhump (approved):

jhump:
@emcfarlane, is there a particular reason this PR is still a draft? If not, I think after you update/rebase it, we can merge it.

PR #935, opened 2024-10-08
Expand test cases for gRPC-Web Trailers to servers-streaming RPCs

timostamm:
We noticed a bug in the gRPC-Web transport from `@connectrpc/connect-web`, where a trailers-only response for a _server-streaming_ RPC failed to raise an error as expected.

I think it makes sense to expand the test cases to cover server-streaming in addition to unary.

### DIFF
--- a/internal/app/connectconformance/testsuites/data/grpc_web_client_trailers.yaml
+++ b/internal/app/connectconformance/testsuites/data/grpc_web_client_trailers.yaml
@@ -10,9 +10,9 @@ relevantCodecs:
 # cover unary and servers-streaming RPCs - the two types that are supported in
 # web browsers.
 testCases:
-  # Trailers and status are in body (no other response messages)
+  # Unary: Trailers and status are in body (no other response messages)
   - request:
-      testName: trailers-in-body/expected
+      testName: unary/trailers-in-body/expected
       streamType: STREAM_TYPE_UNARY
       requestMessages:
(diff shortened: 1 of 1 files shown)


internal/app/connectconformance/testsuites/data/grpc_web_client_trailers.yaml
```diff
@@ -143,3 +145,139 @@ testCases:
       responseTrailers:
         - name: x-custom-trailer
           value: [ "bing", "quuz" ]
+
+  # Server-streaming: Trailers and status are in body (no other response messages)
+  - request:
+      testName: server-stream/trailers-in-body/expected
```
jhump, line 151: For increased clarity of test case names (and more consistency/uniformity), could you rename the existing ones to have a `"unary/"` prefix?

timostamm: Are we okay with potentially breaking known-failing lists?

jhump: Yes.
https://github.com/connectrpc/conformance?tab=readme-ov-file#status-stable

timostamm: Excellent. Done in https://github.com/connectrpc/conformance/pull/935/commits/911c055d30ee47dbb2800ede795057e981b7aa25.

PR #1028, opened 2025-08-27
Fix flaky tests: add an interceptor to be smarter about error code attribution

jhump:
I **think** this change will prevent the semi-regular occurrence of these errors: https://github.com/connectrpc/conformance/actions/runs/17268607815/job/49006867868?pr=1026

In the "context" package, setting a deadline uses a `time.AfterFunc` to create a timer which will cancel the context, asynchronously, when the deadline is reached. But the implementation of `context.Err()` does not actually check the deadline, so it's possible for it to return `nil` after the deadline is passed, because scheduling and execution of the timer function is happening concurrently (and is racing with other code that may have also observed that the deadline has passed).

IIRC, this only really happens with the grpc-go server implementation. That implementation immediately cancels the HTTP/2 stream when it sees the deadline is reached. So what I **think** is happening is that the server canceling the stream and the client then noticing that and attributing a "canceled" RPC code is all happening concurrently with the timer function. It's a race. So sometimes (perhaps surprisingly) the timer function loses the race. So when the client is trying to attribute the RPC code, `ctx.Err()` still returns `nil`. So we fail to tag it as "deadline exceeded" at that point. And we then subsequently tag it as "canceled", inside connect-go's `wrapIfRSTError`.

This change tries to ascertain if the above occurred and then reattribute the error code to "deadline exceeded". To avoid racing with the timer function, it does not examine `ctx.Err()` but instead directly examines the `ctx.Deadline()` and compares it to `time.Now()`.

If this **does** work, I will next commit a similar change directly to `connect.wrapIfRSTError` so that all connect-go clients get this improvement to RPC code attribution. But for now, this seemed easier to try out directly in this repo (without having to update and release connect-go before being able to tell if it made a difference in the tests here).


doriable (approved):
Needed a quick refresh on `wrapIfRSTError`, haha, lgtm!

PR #1104, opened 2026-06-17
Add Connect client-stream test for missing EndStreamResponse

iainmcgin:
A Connect streaming response must end with an `EndStreamResponse` envelope (the `flags & 0b10` frame). The Connect Unexpected Responses suite already covers a client-stream response with **no** data message (`client-stream/ok-but-no-response`) and with **multiple** data messages (`client-stream/multiple-responses`), and the gRPC suite covers the analogous "body present but trailers missing" case (`trailers-only/ignore-header-if-body-present`). There is no Connect-protocol case for a response that delivers the single data message and then closes the body without the end-stream envelope.

A client that doesn't check for the terminus will report such a truncated response as success-with-empty-trailers, which is indistinguishable from a complete response. (We hit this in `connect-rust` — anthropics/connect-rust#163 — and noticed the conformance suite didn't catch it.)

## Expected error code

`CODE_INTERNAL` is the primary expectation: it's what `connect-go` reports (the unexpected-EOF path in `connectStreamingClientConn.Receive` wraps `io.ErrUnexpectedEOF` as `CodeInternal`, which `receiveUnaryMessage` surfaces on the second receive). `CODE_UNKNOWN` and `CODE_UNAVAILABLE` are allowed as alternatives, for the same reason the gRPC analog allows `CODE_UNKNOWN` — the spec doesn't mandate a code, and at least one implementation classifies mid-body truncation as a transport-level (`unavailable`) condition. Happy to narrow this if you'd rather pin a single code.

## Verification

Built `connectconformance` and `referenceclient` from this branch and ran the new case:

```
$ connectconformance --conf ./testing/reference-impls-config.yaml --mode client \
    --run '**/client-stream/no-end-stream' -- referenceclient
Total cases: 5
5 passed, 0 failed
```

Full reference-client suite: 15701 cases, all pass (modulo two pre-existing `Duplicate Metadata/.../bidi-stream/half-duplex/error-with-responses` flakes against the `(grpc server impl)` config that pass on rerun and are unrelated to this change).

## Scope

This PR adds the **client-stream** case only. Server-stream and bidi-stream have the same gap, but the expected response there has to decide whether the data message is yielded to the application before the error fires (most implementations yield it; an implementation that requires the end-stream frame before surfacing anything would not). That's a separate discussion — happy to open it as a follow-up issue if this one lands.

Per CONTRIBUTING.md I'm happy to file an issue first if you'd prefer to discuss the shape before reviewing; opening as a draft so the diff is concrete.

### DIFF
--- a/internal/app/connectconformance/testsuites/data/connect_client_unexpected.yaml
+++ b/internal/app/connectconformance/testsuites/data/connect_client_unexpected.yaml
@@ -100,15 +100,70 @@ testCases:
                     payload:
                       binary_message:
                         "@type": "type.googleapis.com/connectrpc.conformance.v1.ClientStreamResponse"
+    # Per the gRPC status-code guidance, "error parsing returned status" maps
+    # to UNKNOWN: the HTTP body completed cleanly but the protocol-level
+    # terminus is missing.
     otherAllowedErrorCodes:
-      # The spec doesn't mandate a code: connect-go reports `internal`
-      # (unexpected EOF); other implementations report `unavailable` (treats
-      # mid-body truncation as a transport-level condition) or `unknown`.
-      - CODE_UNKNOWN
-      - CODE_UNAVAILABLE
+      # TODO: connect-go currently reports INTERNAL via the unexpected-EOF
+      # path; once it returns UNKNOWN this allowance can be removed.
+      - CODE_INTERNAL
     expectedResponse:
       error:
-        code: CODE_INTERNAL
+        code: CODE_UNKNOWN
+  - request:
+      testName: server-stream/no-end-stream
+      streamType: STREAM_TYPE_SERVER_STREAM
+      requestMessages:
+        - "@type": type.googleapis.com/connectrpc.conformance.v1.ServerStreamRequest
+          responseDefinition:
+            rawResponse:
+              status_code: 200
+              headers:
+                - name: content-type
+                  value: [ "application/connect+proto" ]
+              stream:
+                items:
+                  - flags: 0
+                    payload:
+                      binary_message:
+                        "@type": "type.googleapis.com/connectrpc.conformance.v1.ServerStreamResponse"
+    otherAllowedErrorCodes:
+      # TODO: see client-stream/no-end-stream above.
+      - CODE_INTERNAL
+    expectedResponse:
+      # The data message is well-formed and is yielded to the application
+      # before the missing-end-stream error fires on the next receive.
+      payloads:
+        - { }
+      error:
+        code: CODE_UNKNOWN
+  - request:
+      # Note there is no 'full-duplex' test since that is logically
+      # equivalent to this test and is therefore covered here.
+      testName: bidi-stream/half-duplex/no-end-stream
+      streamType: STREAM_TYPE_HALF_DUPLEX_BIDI_STREAM
+      requestMessages:
+        - "@type": type.googleapis.com/connectrpc.conformance.v1.BidiStreamRequest
+          responseDefinition:
+            rawResponse:
+              status_code: 200
+              headers:
+                - name: content-type
+                  value: [ "application/connect+proto" ]
+              stream:
+                items:
+                  - flags: 0
+                    payload:
+                      binary_message:
+                        "@type": "type.googleapis.com/connectrpc.conformance.v1.BidiStreamResponse"
+    otherAllowedErrorCodes:
+      # TODO: see client-stream/no-end-stream above.
+      - CODE_INTERNAL
+    expectedResponse:
+      payloads:
+        - { }
+      error:
+        code: CODE_UNKNOWN
 
   - request:
       testName: unexpected-content-type
--- a/internal/app/connectconformance/testsuites/data/connect_client_unexpected.yaml
+++ b/internal/app/connectconformance/testsuites/data/connect_client_unexpected.yaml
@@ -79,6 +79,36 @@ testCases:
     expectedResponse:
       error:
         code: CODE_UNIMPLEMENTED
+  - request:
+      # Streaming responses must end with an EndStreamResponse envelope.
+      # A response that delivers the data message and then closes the body
+      # without one is truncated, not complete, and clients should not report
+      # it as success-with-empty-trailers.
+      testName: client-stream/no-end-stream
+      streamType: STREAM_TYPE_CLIENT_STREAM
+      requestMessages:
+        - "@type": type.googleapis.com/connectrpc.conformance.v1.ClientStreamRequest
+          responseDefinition:
+            rawResponse:
+              status_code: 200
+              headers:
+                - name: content-type
+                  value: [ "application/connect+proto" ]
+              stream:
+                items:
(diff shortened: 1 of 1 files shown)


internal/app/connectconformance/testsuites/data/connect_client_unexpected.yaml
```diff
@@ -79,6 +79,36 @@ testCases:
...
+                        "@type": "type.googleapis.com/connectrpc.conformance.v1.ClientStreamResponse"
+    otherAllowedErrorCodes:
+      # The spec doesn't mandate a code: connect-go reports `internal`
+      # (unexpected EOF); other implementations report `unavailable` (treats
+      # mid-body truncation as a transport-level condition) or `unknown`.
+      - CODE_UNKNOWN
+      - CODE_UNAVAILABLE
+    expectedResponse:
+      error:
+        code: CODE_INTERNAL
```
jhump: We follow gRPC's definitions for when to use what codes for things like this. That way we can use the same error code regardless of protocol and still conform to gRPC's public docs.

In this particular case, the table to consult is here:
https://grpc.github.io/grpc/core/md_doc_statuscodes.html

While there is not an entry that straight-forwardly describes exactly this scenario, there are two that match it most closely:

| Case                                                                                                                 | Code    | Generated by |
|----------------------------------------------------------------------------------------------------------------------|---------|--------------|
| Server side application throws an exception or does something other than returning a Status code to terminate an RPC | Unknown | server       |
| Error parsing returned status                                                                                        | Unknown | client       |

In this case, that first one describes the issue most closely, but it also says it is generated by the server (so it's what a server framework should do if the handler fails without returning a status).

The second row is the one that most closely matches on the server side.

So I think we should require `Unknown`. Since the actual conformance servers use connect-go and connect-go returns `Internal`, we can keep `Internal` as another allowed code, but with a TODO comment that we should really only allow `Unknown`, but need some changes in the conformance implementation's dependencies before we can make that true.

So, TLDR, let's drop "unavailable" and then swap "unknown" and "internal" between the two locations.

jhump (commented):
Great addition!

We definitely need coverage of the server and bidi streams, too. But we _also_ need coverage of all four operation kinds with grpc and grpc-web protocols (unary RPCs for those protocols use the same streaming shape as non-unary RPCs).

Would you mind adding the server and bidi cases to this pull request? I can add an issue for the grpc and grpc-web protocols as a follow-up.

Thansk!

iainmcgin:
[claude code]

Thanks for the review and the pointer to the gRPC status-code table — that reasoning makes sense. Pushed `9364b3d` which:

- swaps the codes (`CODE_UNKNOWN` primary, `CODE_INTERNAL` in `otherAllowedErrorCodes` with a TODO referencing connect-go's unexpected-EOF path), and drops `CODE_UNAVAILABLE`;
- adds `server-stream/no-end-stream` and `bidi-stream/half-duplex/no-end-stream`. Those yield the well-formed data message before the missing-end-stream error fires on the next receive, so their `expectedResponse` carries one empty payload alongside the error.

Verified against this branch's runner: connect-go reference client 15/15 on `**/no-end-stream` and 15711/15711 full suite. I've also opened anthropics/connect-rust#168 to switch our client from `unavailable` → `unknown` so it conforms.

Happy to take a stab at the gRPC/gRPC-Web cases too if the follow-up issue would help.

internal/app/connectconformance/testsuites/data/connect_client_unexpected.yaml
iainmcgin: [claude code] Done in 9364b3d — `CODE_UNKNOWN` primary, `CODE_INTERNAL` in `otherAllowedErrorCodes` with a TODO, `CODE_UNAVAILABLE` dropped.

jhump:
@iainmcgin, I think I may have made the wrong call. I'm poking around, and other grpc impls treat this as internal. Also, the actual implementation in connect-go has comments reasoning about the difference and only use "unknown" if the final message has an unknown status code or is missing one. It justifies "internal" as similar to other wire-level issues that align with different entries in that table:

| Case                                                                         | Code     | Generated by |
|------------------------------------------------------------------------------|----------|--------------|
| Could not decompress, but compression algorithm supported (Server -> Client) | Internal | client       |
| Error parsing response proto                                                 | Internal | client       |

So... I guess maybe being lenient as to which code to expect is appropriate since the specs and documentation aren't sufficiently clear. To that end, maybe swap them back, so we "prefer" internal but accept unknown, and remove the TODO.

Sorry for flip-flopping.

> Those yield the well-formed data message before the missing-end-stream error fires on the next receive, so their expectedResponse carries one empty payload alongside the error.

I think this is the right call. The stream impl should deliver messages as they arrive and only return an error after all valid frames have been processed. Impls shouldn't really be buffering or waiting for an end-stream since that might not arrive (for long-lived streams with infrequent messages), so I think it's appropriate to require that impls return the valid messages before returning the error.

jhump:
> I've also opened https://github.com/anthropics/connect-rust/pull/168 to switch our client from unavailable → unknown so it conforms.

Sorry if I've just invalidated that with my last comments. I should have done a little bit more research before my first review reply.

jhump:
> Happy to take a stab at the gRPC/gRPC-Web cases too if the follow-up issue would help.

The grpc-web ones will look exactly the same except for different content-type.

The grpc ones will differ because grpc does not use an end-stream message. It instead crams everything into actual HTTP trailers. I think it should still be straight-forward to map though -- it would be a stream that has no trailers at all. Looks like there are likely some other cases to backfill for grpc, too, like trailers with an invalid/unparseable status code resulting in "unknown". If you want to tackle those, let me know and I'll hold off. But let's do so in a follow-up branch, not in this one.

iainmcgin:
> > I've also opened [anthropics/connect-rust#168](https://github.com/anthropics/connect-rust/pull/168) to switch our client from unavailable → unknown so it conforms.
> 
> Sorry if I've just invalidated that with my last comments. I should have done a little bit more research before my first review reply.

No worries - I've parked that PR until we land this now.

iainmcgin:
[claude code]

No worries on the flip-flop — better to pin it correctly now. Pushed `817554c`:

- swaps all three `no-end-stream` cases (client / server / bidi) to **prefer `CODE_INTERNAL`** with `CODE_UNKNOWN` in `otherAllowedErrorCodes`, and
- drops the TODO that anticipated connect-go moving to `UNKNOWN`. The comment now explains the INTERNAL rationale: a missing end-stream terminus is a wire-level error in the same family as a failed decompression or an unparseable response message, and the spec doesn't mandate a code, so `UNKNOWN` is accepted too.

Verified against this branch's runner: connect-go reference client passes **15/15** on `**/no-end-stream`.

I'll update anthropics/connect-rust#168 to emit `internal` (matching connect-go) once this lands. Happy to take the gRPC / gRPC-Web backfill — including the unparseable-`grpc-status` → `unknown` case you mentioned — in a follow-up branch.

jhump (approved):

PR #1123, opened 2026-08-01
Add gRPC and gRPC-Web coverage for missing terminal status

### LINKED ISSUE #1115: Add gRPC and gRPC-Web coverage for missing terminal status
Following up on #1104: the review called for missing-terminal-status coverage across all four gRPC and gRPC-Web operation kinds. I’m happy to take that on.

My read is that unary/client-streaming should fail without a successful payload, while server-streaming/half-duplex bidi should deliver the valid message, then fail. I’d expect `CODE_INTERNAL`, allowing `CODE_UNKNOWN`.

One question before I put up the PR: grpc-web 1.5.0 and 2.0.2 treat missing trailers as success for unary/server-streaming. Should those be narrow known failures, or should the adapter synthesize the error?

EffortlessSteven:
## What this does

The client suites did not cover the full gRPC and gRPC-Web operation matrix for a valid response message followed by EOF without terminal `grpc-status`.

**Change:** add unary, client-streaming, server-streaming, and half-duplex bidi cases for both protocols. Unary and client-streaming must fail without returning a payload; server-streaming and half-duplex bidi must deliver the valid payload before the terminal error. The browser adapter is unchanged; grpc-web's unary and server-streaming behavior is recorded as two narrow known failures.

## Verification

| Check | Result |
| --- | --- |
| reference client | 4 operations × 7 server configs, all passed |
| grpc-go client 1.82.1 | 4 operations × 1 config, all passed |
| grpc-web client 1.5.0 | 2 advertised operations × 1 config: 2 selected, 2 expected failures, 0 unexpected on Node 22 |
| known-failure scope | the 2 patterns match exactly those 2 cases; the divergence reproduces on grpc-web 2.0.2 |
| negative controls | completing the response or omitting the expected payload made representative cases fail |
| repository gates | `make`, `make checkgenerate`, and `git diff --check` clean in a Linux-native checkout |

## Review map

- `internal/app/connectconformance/testsuites/data/grpc_client_unexpected.yaml`: the four native gRPC cases, terminating with no HTTP trailers.
- `internal/app/connectconformance/testsuites/data/grpc_web_client_unexpected.yaml`: the same operation matrix, omitting the in-body trailers frame.
- `testing/grpcwebclient-known-failing.txt`: records the official grpc-web divergence for unary and server-streaming, the only operation kinds the client advertises.

Closes #1115.


testing/grpcwebclient-known-failing.txt
```diff
@@ -20,6 +20,15 @@ gRPC-Web Unexpected Responses/**/unexpected-compression
 gRPC-Web Unexpected Responses/**/multiple-responses
 gRPC-Web Unexpected Responses/**/ok-but-no-response
 
+# When the HTTP body ends without a terminal trailers frame, the official
+# grpc-web client completes both RPCs without an error. The unary call also
+# returns a payload even though the expected result has none.
+# Keep these cases active as narrow known failures. The entries become stale
+# only when the cases pass their full expectations, including error code and
+# payload count.
```
jhump: The first two sentences look good and are sufficient. These next two sentences are superfluous and read awkwardly IMO. Let's drop them.
```suggestion
```

jhump (commented):
LGTM. Assuming CI checks out (which I assume it will), the only thing I'd want to tweak is the comment in the known-failing file for grpc-web.

testing/grpcwebclient-known-failing.txt
EffortlessSteven: Done, thanks!

